PR #7, opened 2015-11-25
use new carddav endpoint in core

DeepDiver1975:


DeepDiver1975:
@irgendwie works - let's merge this

PR #9, opened 2015-11-25
layout for list of contacts, using core classes ported from Mail app

DeepDiver1975:


jancborchardt:
Do your own branches homie

PR #18, opened 2016-02-10
Addressbook sharing

tomneedham:
TODO:
- [x] Interpret shares on load of the AddressBook model
- [x] Tweak some css in the UI to make it pretty


DeepDiver1975:
ouch - looks :boom:

tomneedham:
some squashing needed... ;)

tomneedham:
@DeepDiver1975 Broken for you?

DeepDiver1975:
Looks like a rebase is needed ... Toooo many commits

tomneedham:
Mmm clean now :)

bower.json
```diff
@@ -22,6 +22,9 @@
     "angular-uuid4": "~0.3.1",
     "angular-cache": "~4.5.0",
     "angular-route": "^1.5.0",
-    "vcard": "~0.2.3"
+    "vcard": "~0.2.3",
+    "dav": "git@github.com:irgendwie/dav.git#d33ecad1e64202ec5b35b6911708d1253064b5a7",
+    "vcard": "~0.2.3",
```
Henni: We removed the dav library from bower in one of the last PRs.
Additionally vcard seems to be twice in this list.

js/services/addressBook_service.js
```diff
@@ -72,7 +72,111 @@ app.factory('AddressBookService', ['DavClient', 'DavService', 'SettingsService',
...
+			oSet.appendChild(oSummary);
+
+			if (writable) {
+				var oRW = xmlDoc.createElement('o:read-write');
+				oSet.appendChild(oRW);
+			}
+
+			var body = oShare.outerHTML;
+
+			DavClient.xhr.send(
```
Henni: I believe this is missing a `return`. I'm not sure if https://github.com/owncloud/contactsrework/pull/18/files#diff-dbaa024398385bf6cffe22c422471ff0R96 works without this.

js/public/script.js
```diff
@@ -163,12 +264,66 @@ app.factory('AddressBook', function()
...
+		}
+
+		var owner = props['{DAV:}owner'];
+		if (typeof owner !== 'undefined' && owner.length !== 0) {
+			owner = owner[0].textContent.slice(0, -1);
+			if (owner.startsWith('/remote.php/dav/principals/users/')) {
+				this._properties.owner = owner.substr(33);
+			}
+		}
+		*/
```
Henni: I know that this is still work in progress... But just a short reminder to remove this code comment block before merging.

templates/main.php
```diff
@@ -12,8 +12,13 @@
 // compiled version of app javascript
 script('contactsrework', 'public/script');
 
+script('contactsrework', 'vendor/angular-bootstrap/ui-bootstrap.min');
+script('contactsrework', 'vendor/angular-bootstrap/ui-bootstrap-tpls.min');
+
 // all styles
 style('contactsrework', 'public/style');
+style('contactsrework', 'addressbooklist');
```
Henni: Please move the styles to `public/style.css`.
We decided to keep all styles in a single file and remove any modularization or build steps for now.

js/public/script.js
tomneedham: Yup this is my thinking out loud of how we interpret shares on the initial PROPFIND. @DeepDiver1975 is implementing it in the dav library tomorrow and then we can use this code.

templates/main.php
tomneedham: Cool shall do. Again this was mainly whilst stuff got worked out.

DeepDiver1975:
@Henni @tomneedham @jancborchardt sharees are now read from the propfind response - have fun!

js/models/addressBook_model.js
```diff
@@ -13,9 +13,56 @@ app.factory('AddressBook', function()
...
+
+			sharedWith: {
+				users: [],
+				groups: []
 			}
 
 		});
 		angular.extend(this, data);
+		angular.extend(this, {
+			owner: data.url.split('/').slice(-3, -2)[0]
```
tomneedham: @DeepDiver1975 Is this ok (with regards to how I have found the owner?). Calendar uses the share prop that you just implemented, is that safer?

tomneedham:
@DeepDiver1975 We get a load of lines in the logs that 'principals' sharebackend is not found because of https://github.com/owncloud/core/blob/master/apps/files_sharing/api/sharees.php#L445. Should we add a dummy backend to return no to this and keep the log quiet?

DeepDiver1975:
> @DeepDiver1975 We get a load of lines in the logs that 'principals' sharebackend is not found because of https://github.com/owncloud/core/blob/master/apps/files_sharing/api/sharees.php#L445. Should we add a dummy backend to return no to this and keep the log quiet?

or we simply use file for now ....

DeepDiver1975:
I'll rebased this beast ....

js/models/addressBook_model.js
DeepDiver1975: I have to double check this ...

Actually I wanted to ask you what for the owner is used.
Best would be to get it from the profind response ...

jancborchardt:
Ok … not sure how I can test this. The UI doesn’t seem to change at all, and I assume there should be something in the settings area?

DeepDiver1975:
Sharing by admin:
bildschirmfoto von 2016-02-24 15-54-49
bildschirmfoto von 2016-02-24 15-55-24
bildschirmfoto von 2016-02-24 15-55-33


As user01:
bildschirmfoto von 2016-02-24 15-56-03

jancborchardt:
How do I get from screenshot 1 to 2? I assume clicking the address book – doesn’t work though here.

Just to be sure, the element looks like this:
```
<li class="ng-scope ng-binding ng-isolate-scope"
  ng-repeat="addressBook in ctrl.addressBooks"
  addressbook="" data="addressBook">
    testAddr
</li>
```

DeepDiver1975:
> How do I get from screenshot 1 to 2? I assume clicking the address book – doesn’t work though here.

there is a share icon (which is not in the screenshot) ... just a sec

DeepDiver1975:
bildschirmvideo von 24 02 2016 16_37_59

jancborchardt:
For the record: doesn’t work with Firefox, only Chrome.

DeepDiver1975:
> For the record: doesn’t work with Firefox, only Chrome.

@tpn please take care - THX

DeepDiver1975:
sharing works - the shared addressbook appears on the sharee - but no contacts are loaded - I suggest to take care of this in a followup pr

DeepDiver1975:
> but no contacts are loaded

this is because of the enabled logic - https://github.com/owncloud/contactsrework/blob/master/js/services/addressBook_service.js#L35

DeepDiver1975:
:+1:

DeepDiver1975:
let get this in @jancborchardt

js/models/addressBook_model.js
tomneedham: Yup this is how calednar does it: https://github.com/owncloud/calendar-rework/blob/master/js/app/models/calendarmodel.js#L107

tomneedham:
> For the record: doesn’t work with Firefox, only Chrome.

Yup I see this too, created https://github.com/owncloud/contactsrework/issues/54

jancborchardt:
I tested this in Chrome: I do get an address book for my other user called `testAddr(jan)` but nothing happens … no contacts displayed by default, nothing changes when I click on »All« or on that address book. cc @DeepDiver1975 
Maybe best to merge anyway and fix stuff going forward?

@tomneedham some additional things to fix, although in a future PR:
- [ ] clicking on the share icon should not only show the share box, but also focus it with the cursor
- [ ] the suggestion dropdown should overlay the content below, not move the whole thing up like now
- [ ] remove the dash behind the name in the list

DeepDiver1975:
> @tomneedham some additional things to fix, although in a future PR:
> 
>     clicking on the share icon should not only show the share box, but also focus it with the cursor
>     the suggestion dropdown should overlay the content below, not move the whole thing up like now
>     remove the dash behind the name in the list

let's add this to #54

PR #19, opened 2016-02-11
Show delete immediately

Henni:
Besides implementing the delete functionality I cleaned up the code and updated 3rdparty libraries (including dav.js).


js/models/contact_model.js
```diff
@@ -69,7 +69,10 @@ app.factory('Contact', [ '$filter', function($filter) {
 
 			setETag: function(etag) {
 				this.data.etag = etag;
-				this.data.props.getetag = etag;
+			},
+
+			setUrl: function(uid) {
+				this.data.url = "/remote.php/dav/addressbooks/users/hendrik/testAddr/"+ uid +".vcf";
```
DeepDiver1975: :ghost:

Henni: jej thanks...

Henni: @DeepDiver1975 fixed

js/models/contact_model.js
```diff
@@ -69,7 +69,10 @@ app.factory('Contact', [ '$filter', function($filter) {
 
 			setETag: function(etag) {
 				this.data.etag = etag;
-				this.data.props.getetag = etag;
+			},
+
+			setUrl: function(addressBook, uid) {
+				this.data.url = OC.linkToRemoteBase('dav/addressbooks/users/') + OC.getCurrentUser().uid + '/' + addressBook.displayName + '/' + uid + ".vcf";
```
DeepDiver1975: displayname? should be uri

Henni: `displayName` works as well because I'm building the url myself. But you are right using the url simplifies this step

js/models/contact_model.js
```diff
@@ -69,7 +69,10 @@ app.factory('Contact', [ '$filter', function($filter) {
 
 			setETag: function(etag) {
 				this.data.etag = etag;
-				this.data.props.getetag = etag;
+			},
+
+			setUrl: function(addressBook, uid) {
+				this.data.url = addressBook.url + uid + ".vcf";
```
Henni: @DeepDiver1975 happy with this solution?

DeepDiver1975:
- [ ] deleting a contact does not update the right side bar
- [ ] delete of a fresh created contacts is not possible (because on select it is not loaded on the right)

Henni:
I'll fix 1, but I can't reproduce 2.

DeepDiver1975:
> but I can't reproduce 2.

selection only works if you click on the name - not on the row (but the high lighting works)

Henni:
This is a design issue, @jancborchardt wanted to look into.

DeepDiver1975:
> This is a design issue, @jancborchardt wanted to look into.

isn't this more a question on which element the change event is setup?

from an UX pov a contacts should be selected when clicking the row (anywhere) and not only on the name of the contact

Henni:
Yes we will change it, so that the complete row triggers the details. But I would like to wait until we have a working UI.

DeepDiver1975:
ok

DeepDiver1975:
:+1:

PR #20, opened 2016-02-11
Debounce updates

Henni:
This PR limits updates to twice a second.

Currently the input itself is debounced. We might move this to service later on, so that the changes propagate through the client, but not to the server.

### DIFF
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -1,11 +1,11 @@
 <header class="contactdetails__header"
 		ng-style="{'background-color': (ctrl.contact.uid() | contactColor)}">
 	<h2 style="font-size: 1.7em; margin-left: 1em;">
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
- [ ] on change of email there is never an update triggered

DeepDiver1975:
- [ ] (unrelated to this pr) right sidebar content only changes if clicking on the name

Henni:
> on change of email there is never an update triggered

this isn't implemented yet, but isn't part of this PR

> right sidebar content only changes if clicking on the name

what do you mean by this?

DeepDiver1975:
> what do you mean by this?

https://github.com/owncloud/contactsrework/pull/19#issuecomment-182913877 :wink:

DeepDiver1975:
:+1:

PR #22, opened 2016-02-12
Stupid hack to get customer property 'groups' into the propfind

DeepDiver1975:
requires https://github.com/owncloud/core/pull/22294

@hechi @irgendwie so this is what we need - how do we get this properly integrated into the dav client lib?


DeepDiver1975:
wtf? all lines changed?

DeepDiver1975:
as soon as I touch that file on linux I get a hugh change set - maybe line endings?

DeepDiver1975:
and the client code cannot properly work with the group response

```xml
<oc:groups>
    <oc:group>A</oc:group>
    <oc:group>B</oc:group>
</oc:groups>
```

is a concatinated string on the client

PR #23, opened 2016-02-12
layout for list of contacts, using core classes ported from Mail app

jancborchardt:
First very basic version to make it exactly like the Mail app. In the future we can tweak it, but should aim to develop Mail and Contacts the same in that regard so we can move the classes to core. One thing for example would be to make the avatars bigger.

capture du 2016-02-12 12-10-07

Please review @Henni @owncloud/designers @DeepDiver1975

### DIFF
--- a/css/contactlist.css
+++ b/css/contactlist.css
@@ -1,24 +1,3 @@
-#contactstable {
-	width: 100%;
-}
-
-#contactstable th {
-	color: #999;
-}
-
-#contactstable th,
-#contactstable td {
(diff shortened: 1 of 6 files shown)


Henni:
The star looks a bit off, especially without an email, but I like it otherwise. :+1:

jancborchardt:
In most cases there's probably another line, like email, phone, or company. So I suppose thus will be less of a problem.

But as said, this is only the first iteration. :)

PR #28, opened 2016-02-12
Show groups on the left side bar

DeepDiver1975:
bildschirmfoto von 2016-02-15 16-10-43


Henni:
Should this still be part of 1.0?

DeepDiver1975:
> Should this still be part of 1.0?

no - I already moved it to 2.0 :wink:

PR #29, opened 2016-02-12
The full row is clickable

### LINKED ISSUE #26: Whole row should be clickable
Instead of only the name / first line, the whole `app-content-list-item` row should be clickable to open the contact.

DeepDiver1975:
fixes #26


jancborchardt:
Fixed the a element not taking up the whole row. Works now! :+1:

PR #32, opened 2016-02-13
Remove css task from gulp

Henni:
Removes the css task from gulp to simplify contributing.
Because of this change all styles should be included in styles.css

cc @jancborchardt


jancborchardt:
Yup, thanks a lot! :) Because also most of the styles that we use should be served from core anyway.

PR #33, opened 2016-02-15
Remove ui-router

Henni:
A bit of refactoring to remove angular ui-router and replace it with angular router.


DeepDiver1975:
tested - still works :+1:

PR #40, opened 2016-02-18
Add CSRF token to each request

LukasReschke:
Big hack. But I have no idea about AngularJS :speak_no_evil:

Required once https://github.com/owncloud/core/pull/22424 is in …

@Henni Any idea how to solve this cleaner? All hijacking permitted :wink:

### DIFF
--- a/js/dav/dav.js
+++ b/js/dav/dav.js
@@ -3106,6 +3106,7 @@ var Basic = (function (_Transport) {
               result = undefined;
               context$3$0.prev = 9;
               context$3$0.next = 12;
+              xhr.setRequestHeader('requesttoken', oc_requesttoken);
               return xhr.send(request.requestData);
 
             case 12:


DeepDiver1975:
https://github.com/owncloud/calendar-rework/blob/master/js/config/config.js#L28

LukasReschke:
I think `httpProvider` just applies to the `$http` service of AngularJS: https://docs.angularjs.org/api/ng/service/$http

Since the library does not use `$http` it won't have any effect. That's also why https://github.com/owncloud/calendar-rework/pull/218 is there. Then again, I'm far away from being a AngularJS expert. At least it didn't work before. :hear_no_evil:

DeepDiver1975:
needs rebase ....

DeepDiver1975:
fine by me :+1:

DeepDiver1975:
oh . wait dav.js is generated ...
let me fix this

DeepDiver1975:
here we go https://github.com/owncloud/contactsrework/pull/43

PR #44, opened 2016-02-18
Don't use wellknown if not needed

### LINKED ISSUE #41: Do not use .well-known/carddav url
For me there is a 404 request to the .well-known/carddav url in the network log:
`http://localhost/.well-known/carddav`

DeepDiver1975:
fixes #41

@nickvergessen @Henni

### DIFF
--- a/js/dav/dav.js
+++ b/js/dav/dav.js
@@ -683,7 +683,7 @@ var serviceDiscovery = _co2['default'].wrap(regeneratorRuntime.mark(function cal
         uri = _url2['default'].format({
           protocol: endpoint.protocol,
           host: endpoint.host,
-          pathname: '/.well-known/' + options.accountType
+          pathname: !options.useProvidedPath ? '/.well-known/' + options.accountType : endpoint.pathname
         });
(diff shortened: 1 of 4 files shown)


Henni:
Looks good to me. Still works and there is no 404 anymore.

DeepDiver1975:
rebased

nickvergessen:
For me it still queries the well-known, because the actual lib patch is missing

nickvergessen:
But then again, the thing only makes a get request, which returns
> This is the WebDAV interface. It can only be accessed by WebDAV clients such as the ownCloud desktop sync client.

So maybe we should just kill that get call completly...

DeepDiver1975:
> But then again, the thing only makes a get request, which returns

grrr - where did it go?

DeepDiver1975:
> For me it still queries the well-known, because the actual lib patch is missing

fixed - please test again - thx

DeepDiver1975:
@nickvergessen THX

DeepDiver1975:
rebased again

irgendwie:
LGTM :+1:

Issue #36, opened 2016-02-17
Mobile styles


jancborchardt:
We need to copy over the mobile styles from the Mail app so contacts properly works on small screens.

tomneedham:
What's the plan for the layout, right now I get this: 
![screen shot 2016-02-18 at 14 24 59](https://cloud.githubusercontent.com/assets/660805/13145978/75f519ae-d64b-11e5-8204-0fa421e60294.png)

jancborchardt:
@tomneedham if you want to take a stab at this, check out the Mail app. Essentially you can copy the mobile.css from there, of course with class name adjustments.

jancborchardt:
What should happen is:
- by default, the contact list shows, with the sidebar button on the top left (like now)
- tapping a contact opens the detail view
- the detail view looks like now, but has a back button to go back to the contact list

tomneedham:
** tries mail app _again_ ** ;-)

tomneedham:
See IRC: #owncloud-contacts please :)

skjnldsv:
Should I suggest this? :)
![contacts owncloud](https://cloud.githubusercontent.com/assets/14975046/14103142/e64ceed0-f59e-11e5-8a03-d08b50d29100.png) ![contacts owncloud2](https://cloud.githubusercontent.com/assets/14975046/14103141/e64c5e02-f59e-11e5-8e2f-e5ce3edf3b88.png)

Instead of this:
![contacts owncloud3](https://cloud.githubusercontent.com/assets/14975046/14103146/e95b8fd2-f59e-11e5-8979-11911c609de3.png)

skjnldsv:
Okay, just saw that's exaclty what  @tomneedham  did in the PR, good job! ;)

jancborchardt:
@skjnldsv yeah dude, and additionally your replacing of the menu button with the back button in the details view is great! :)

skjnldsv:
Eheh

PR #46, opened 2016-02-18
Add mobile style from Mail app, ref #36

jancborchardt:
What needs to be done in JS:
- by default, the contact list shows, with the sidebar button on the top left (like now)
- tapping a contact opens the detail view
- the detail view looks like now, but has a back button to go back to the contact list on the top left, instead of the sidebar button.

cc @Henni @DeepDiver1975 @tomneedham

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -140,7 +140,7 @@ detailsitem input[type="date"],
 detailsitem textarea,
 .select-addressbook select,
 .add-field {
-	width: 200px;
+	width: 245px;
 }
 
 detailsitem .icon-delete {
@@ -393,6 +393,19 @@ li.addressBook-share-item span.shareeIdentifier {
 	opacity: .5;
 }
 
+ul.addressBookList li {
+    margin-bottom: 5px;
+}
+
+ul.addressBook-share-list {
+    margin-top: 8px;
+    margin-bottom: 12px;
+}
+
+ul.addressBook-share-list li {
+    margin-left: 15px;
+}
+
 #app-navigation .utils .action span {
 	cursor: pointer !important;
 }
@@ -404,14 +417,10 @@ li.addressBook-share-item span.shareeIdentifier {
 	opacity: 1;
 }
 
-
-
-
-
 /* SELECT2 styling - MOVE TO CORE FOR 9.1! */
 
 detailsitem .select2-container {
-	width: 200px;
+	width: 245px;
 }
 .select2-container-multi .select2-choices {
 	border-color: #ddd !important;
@@ -444,4 +453,64 @@ detailsitem .select2-container {
 }
 .select2-results .select2-result-label span {
 	cursor: pointer !important;
+}
+
+/* Mobile width < 768px */
+@media only screen and (max-width: 768px) {
+
+	/* full width for message list on mobile */
+	.app-content-list {
+		width: 100%;
+		background: white;
+		position: relative;
+		z-index: 100;
+	}
+
+	/* overlay message detail on top of message list */
+	.app-content-detail {
+		background: #fff;
+		width: 100%;
+		left: 0;
+		height: 100%;
+		top: 0;
+		box-shadow: 0 0 100px rgba(100, 100, 100, .9);
+		position: absolute;
+	}
+	
+	.wrapper-show:not(.mobile-show),
+	.contacts-list:not(.mobile-show) {
+		display: none;
+	}
+	
+	
+	#app-navigation-toggle.showdetails {
+		transform: translate(-50px, 0);
+	}
+	
+	#app-navigation-toggle-back {
+		position: fixed;
+		display: inline-block !important;
+		top: 45px;
+		left: 0;
+		width: 44px;
+		height: 44px;
+		z-index: 149;
+		background-color: rgba(255, 255, 255, .7);
+		cursor: pointer;
+		opacity: .6;
+		transform: rotate(90deg);
+	}
+
+/* end of media query */
+}
+
+.contact-details-wrapper {
+	position: relative;
+	background: white;
+}
+.wrapper-show {
+	z-index: 101;
+}
+#app-navigation-toggle-back {
+	display: none;
 }
--- a/js/components/contactDetails/contactDetails_controller.js
+++ b/js/components/contactDetails/contactDetails_controller.js
@@ -1,8 +1,19 @@
 angular.module('contactsApp')
-.controller('contactdetailsCtrl', function(ContactService, AddressBookService, vCardPropertiesService, $routeParams, $scope) {
+.controller('contactdetailsCtrl', function(ContactService, AddressBookService, vCardPropertiesService, $route, $routeParams, $scope) {
+
 	var ctrl = this;
 
 	ctrl.loading = true;
+	ctrl.show = false;
+
+	ctrl.clearContact = function() {
+		$route.updateParams({
+			gid: $routeParams.gid,
+			uid: undefined
+		});
+		ctrl.show = false;
+		ctrl.contact = undefined;
+	};
 
 	ctrl.uid = $routeParams.uid;
 	ctrl.t = {
@@ -35,11 +46,16 @@ angular.module('contactsApp')
 
 	ctrl.changeContact = function(uid) {
 		if (typeof uid === 'undefined') {
+			ctrl.show = false;
+			$('#app-navigation-toggle').removeClass('showdetails');
 			return;
 		}
 		ContactService.getById(uid).then(function(contact) {
 			ctrl.contact = contact;
 			ctrl.photo = ctrl.contact.photo();
+			ctrl.show = true;
+			$('#app-navigation-toggle').addClass('showdetails');
+
 			ctrl.addressBook = _.find(ctrl.addressBooks, function(book) {
 				return book.displayName === ctrl.contact.addressBookId;
 			});
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -1,35 +1,38 @@
-<div ng-if="ctrl.contact===undefined && !ctrl.loading">
-	<div id="emptycontent" class="">
-		<div class="icon-contacts-dark"></div>
-		<h2>{{ctrl.t.noContacts}}</h2>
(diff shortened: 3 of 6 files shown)


jancborchardt:
@tomneedham @Henni @irgendwie @owncloud/javascript any JS help to spare? :)

DeepDiver1975:
I suggest to create a new issue for the open todos - we can sort this out in a later stage

DeepDiver1975:
@jancborchardt is the css side done?

jancborchardt:
Normally I would also do it step-by-step, but in this case it’s better to do it all at once. How it is now it looks ugly, but at least it works. With the pull request half-done, there’s basically no way at all to view contact details.

So I would rather do this all at once. And it would actually be fine to do this after 9.0

DeepDiver1975:
> So I would rather do this all at once. And it would actually be fine to do this after 9.0

Okay - lets postpone this until we have the basic functionality there - the details view is prio for now

jancborchardt:
Yep, working on layouting that. Is the avatar coming?

DeepDiver1975:
> Yep, working on layouting that. Is the avatar coming?

https://github.com/owncloud/contactsrework/pull/51#issuecomment-188306187

tomneedham:
Rebased

DeepDiver1975:
> Rebased

again

DeepDiver1975:
@jancborchardt have there been changes to js originally?

jancborchardt:
@DeepDiver1975 what do you exactly mean? No, this pull request atm is pure CSS changes. And as said above until the necessary JS additions are done, we should definitely not merge it, especially not for 9.0.

DeepDiver1975:
I fixed this already - rebase fun :wink:

tomneedham:
Rebased and added some field width fixes. Just looking at what JS work needs to be done here
screen shot 2016-03-10 at 15 52 02
(Maximised field width and aligned delete icons)

tomneedham:
So in principle I have this working by altering the `z-index` on the `app-content-details` to temporarily hide this behind the list. The difficulty is now controlling this property within angular. Atm I am trying a wrapper around the details html which has `ngStyle` for the z index

Henni:
@tomneedham I'm not completely sure what you want to do, but I'd suggest using `ng-class` instead of `ng-style` to avoid inline style.

tomneedham:
I'm on the cusp of a breakthrough ;)

tomneedham:
@Henni Trying to make the 'back' button. How can I 'clear' the uid from the url ?

tomneedham:
So I've added a back button which shows on mobile. Unfortunately it's a bit of a hack at the moment as it just removes the view and doesn't change the url like it should. This has the side effect that if you select the same contact after clicking back, the url doesnt change so it doesn't run the changeContact code.  @Henni might be able to help here with some ngRoute knowledge.

tomneedham:
@jancborchardt Back button needs styling

js/components/contactDetails/contactDetails_controller.js
```diff
@@ -1,6 +1,13 @@
 app.controller('contactdetailsCtrl', ['ContactService', 'AddressBookService', 'vCardPropertiesService', '$routeParams', '$scope', function(ContactService, AddressBookService, vCardPropertiesService, $routeParams, $scope) {
 	var ctrl = this;
 
+	ctrl.show = false;
+
+	ctrl.clearContact = function() {
+		delete $routeParams.uid;
+		ctrl.show = false;
+	};
+
```
tomneedham, line 17: TODO: Fix this, this was a feeble attempt at clearing the param from the URL. This should be triggered with ng-click on the back button.

Henni, line 17: I'm currently not able to test it, but try setting the uid to `null` instead of deleting it.

tomneedham:
Working nicely now - oh rebasing this is proving fun.....

tomneedham:
@jancborchardt Can you test this please. Works from my POV just need styling on the back button when in mobile viewing a contact

irgendwie:
@tomneedham Nice work there! :+1:
Currently choosing a group leads to first contact being selected (due to #233),
but the contact-list should be shown instead, meaning no contact should be selected.

tomneedham:
Hmm what's the best way to fix this - we could add something like https://github.com/jacopotarantino/angular-match-media to query the screen size and not select the first contact - but is there an easier way here?

tomneedham:
rebased

css/public/style.css
```diff
@@ -438,3 +447,42 @@ detailsitem .select2-container {
...
+}
+
+button.details-back {
+	display: block !important;
+}
+
+/* end of media query */
+}
+
+.contact-details-wrapper {
```
skjnldsv, line 507: We need `height: 100%;` to avoid having the list poping up under the contactDetail: :)
contacts owncloud

templates/contactDetails.html
```diff
@@ -1,35 +1,33 @@
...
-</header>
-<section>
-	<div ng-repeat="prop in ctrl.contact.props | toArray | orderDetailItems:'$key'">
-		<detailsItem ng-repeat="propData in prop" name="prop.$key" data="propData" model="ctrl" index="$index" ng-class="[ 'details-item-' + prop.$key ]"></detailsItem>
-	</div>
-	<div class="select-addressbook" ng-if="ctrl.addressBooks.length > 1">
-		<select ng-model="ctrl.addressBook" ng-change="ctrl.changeAddressBook(ctrl.addressBook)" ng-options="book.displayName for book in ctrl.addressBooks">
+	<div ng-if="ctrl.contact!==undefined">
+	<header class="contactdetails__header" ng-style="{'background-color': (ctrl.contact.uid() | contactColor)}">
+		<button class="details-back" ng-click="ctrl.clearContact()">Back</button>
```
skjnldsv: Can I suggest using this instead?
`<div id="app-navigation-toggle-back" class="details-back icon-download" ng-click="ctrl.clearContact()"></div>`

```
#app-navigation-toggle-back {
    position: fixed;
    display: inline-block !important;
    top: 90px;
    left: 0;
    width: 44px;
    height: 44px;
    z-index: 149;
    background-color: rgba(255, 255, 255, .7);
    cursor: pointer;
    opacity: .6;
    transform: rotate(90deg);
}
```

contacts owncloud

skjnldsv:
@tomneedham for the auto-loading of the first contact, we could set two variables?
1. ctrl.initialshow
2. ctrl.show

It could allow us to differentiate a user selection from a default page initialisation.
Therefore, we should be able to have to different css classes and use the media queries to use or not use the first auto-selection of the first contact for the mobile layout. :)

jancborchardt:
The back button should replace the menu button instead of being below it. Otherwise it’s too complex. :) That’s also how it’s done on Android. Good job!

skjnldsv:
Indeed, and there is an animation (rotation) from the menu icon to the arrow !

jancborchardt:
Yeah – we don’t necessarily need to get so fancy though, especially for performance reasons on mobile. ;)

jancborchardt:
@skjnldsv just let me know as soon as you updated this (also needs a rebase) and I can review. :)

skjnldsv:
Want me to update this branch or fork a new one?

jancborchardt:
Update this one – collaboration is in order when it’s a modification of the same thing. :)

tomneedham:
Thanks @skjnldsv that is a much better idea :)

skjnldsv:
> Hmm what's the best way to fix this - we could add something like https://github.com/jacopotarantino/angular-match-media to query the screen size and not select the first contact - but is there an easier way here?

I'm simply gonna use `$(window).width()>768`, much easier!

skjnldsv:
Rebased and finished.
- Maximise the field width and align delete icons
- Added back button replacing the menu when contact details is opened
- Won't load first contact when mobile layout on

@tomneedham @Henni @irgendwie @DeepDiver1975 @jancborchardt

jancborchardt:
Btw @tomneedham @skjnldsv when multiple people worked on the same pull request, you don’t rebase / squash into one commit of only one of the people. ;) Now it’s very difficult to check the changes. Just a thing for the future.

skjnldsv:
Oh my bad. :/

skjnldsv:
@jancborchardt should we merge?

jancborchardt:
@skjnldsv you need to fix the merge conflicts first :) – another reason why it’s best to have multiple commits, cause now I don’t exactly know where the error lies. ;)

skjnldsv:
@jancborchardt How am I supposed to do that? 
Should i commit again with a fix for conflicts files that include the recent commits?

css/public/style.css
skjnldsv, line 507: Finally no, we need to hide the contactList instead, this trick doesn't work on all devices.

skjnldsv:
Just did a few fixes again, the height:100% wasn't working on some device. This wasn't a good trick. Instead we needed to hide the contactList.
I also fixed an issue which was selecting the first contact when clicking a group.

skjnldsv:
Sh*\* I did a mistake!
I'm lost, I don't know what's going on with all the commit merge errors, and I screwed the rebase.
Shouldn't we create a new fresh PR?

Sorry! :/

EDIT i rolled back my mistake!
@jancborchardt can you merge now?

jancborchardt:
@skjnldsv it still says there are conflicts. Basically you need to do:

```
git checkout master
git pull
git checkout this branch
git rebase master
then fix the conflicts, git add them, and then push here
```

skjnldsv:
Well, I hope I did right.
I'm not familiar enough to know what to do in case of conflicts. When everythings fine, git is ok for me, but at the first sight of an issue, i'm lost :p 
Thank for your help @jancborchardt !

jancborchardt:
@skjnldsv did you check out the mobile view (that is, make your browser window narrow)? For me it stays completely white. Regardless if I size it down from wider or if I refresh when already narrow. Can you look into that? Cause right now mobile is broken. ;)

skjnldsv:
Do you have any contacts?

jancborchardt:
Aaah, now it works. No idea what was the issue. Great stuff! :) :raised_hands:

skjnldsv:
:+1:

PR #55, opened 2016-02-24
Initial setup of l10n

DeepDiver1975:


DeepDiver1975:
bildschirmfoto von 2016-02-24 21-11-09

jancborchardt:
:+1: except for 2 things to grep-change:
- [ ] »Add Contact« should be »Add contact« with lowercase second word
- [ ] »All« should maybe be »All contacts«? Because it looks a bit off otherwise.

PR #60, opened 2016-02-25
Design enhancements

jancborchardt:
Just 2 small enhancements to the detail view header:
- make avatar round, like all others
- add placeholder »Name« to name input so it’s more visible when empty

### DIFF
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -8,7 +8,7 @@
 <header class="contactdetails__header" ng-style="{'background-color': (ctrl.contact.uid() | contactColor)}">
 	<img class="contactdetails__logo avatar" data-ng-src="data:image/png;base64,{{ctrl.photo}}" />
 	<h2>
-		<input type="text" id="details-email" class="contactdetails__name" placeholder="Name" name="email" ng-model="ctrl.contact.fullName" ng-model-options="{ getterSetter: true, debounce: 500 }" ng-change="ctrl.updateContact()" value="" />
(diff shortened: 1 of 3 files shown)


templates/contactDetails.html
```diff
@@ -6,9 +6,9 @@
 </div>
 <div ng-if="ctrl.contact!==undefined">
 <header class="contactdetails__header" ng-style="{'background-color': (ctrl.contact.uid() | contactColor)}">
-	<img class="contactdetails__logo" data-ng-src="data:image/png;base64,{{ctrl.photo}}" />
+	<img class="contactdetails__logo avatar" data-ng-src="data:image/png;base64,{{ctrl.photo}}" />
 	<h2>
-		<input type="text" id="details-email" class="contactdetails__name" name="email" ng-model="ctrl.contact.fullName" ng-model-options="{ getterSetter: true, debounce: 500 }" ng-change="ctrl.updateContact()" value="" />
+		<input type="text" id="details-email" class="contactdetails__name" placeholder="Name" name="email" ng-model="ctrl.contact.fullName" ng-model-options="{ getterSetter: true, debounce: 500 }" ng-change="ctrl.updateContact()" value="" />
```
DeepDiver1975: We need to translate that

DeepDiver1975:
no so cool if the photo is not square

bildschirmfoto von 2016-02-26 07-36-46

jancborchardt:
The photo should always be cut to display square. Or at least we should force it square via CSS. Smth for a different PR though.

DeepDiver1975:
> Or at least we should force it square via CSS

we should do this - pr welcome

Henni:
looks good otherwise :+1:

PR #67, opened 2016-02-29
THIS IS CONTACTS!

DeepDiver1975:


appinfo/info.xml
```diff
@@ -1,12 +1,11 @@
 <?xml version="1.0"?>
 <info>
-    <id>contactsrework</id>
+    <id>contacts</id>
     <name>Contacts</name>
     <description>The new and improved app for your Contacts.</description>
     <licence>AGPL</licence>
     <author>Hendrik Leppelsack</author>
     <version>0.0.1</version>
```
nickvergessen: Must be higher then the old one:
https://github.com/owncloudarchive/contacts/blob/master/appinfo/info.xml#L21

``` xml
<version>0.5.0.0</version>
```

nickvergessen:
```xml
<version>0.5.0.0</version>
```

package.json
```diff
@@ -9,13 +9,13 @@
   "private": true,
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/Henni/contactsrework.git"
+    "url": "git+https://github.com/Henni/contacts.git"
```
nickvergessen: `https://github.com/owncloud/contacts.git`

package.json
```diff
@@ -9,13 +9,13 @@
...
   },
   "license": "AGPL-3.0",
   "bugs": {
-    "url": "https://github.com/Henni/contactsrework/issues"
+    "url": "https://github.com/Henni/contacts/issues"
```
nickvergessen: `https://github.com/owncloud/contacts/issues`

package.json
```diff
@@ -9,13 +9,13 @@
...
   },
-  "homepage": "https://github.com/Henni/contactsrework#readme",
+  "homepage": "https://github.com/Henni/contacts#readme",
```
nickvergessen: `https://github.com/owncloud/contacts#readme`

nickvergessen:
- [x] info.xml has no `<ocsid>168708</ocsid>` entry?

Should add it for updating, or will this break stuff?

DeepDiver1975:
> Should add it for updating, or will this break stuff?

Should we use the same or create a new one? I honestly don't know the impact of the decision.

appinfo/info.xml
DeepDiver1975: Since the new contacts app has no migrations this is pointless from my poc.

But we should still adopt this properly.

Let's name it 1.0.0.0

LukasReschke:
> Should we use the same or create a new one? I honestly don't know the impact of the decision.

If you consider to use the existing appstore entry and want to have automatic updates of the application: Keep the ID of the old app. Just make sure to upload it for 9.0 only to the appstore and don't touch the older builds.

DeepDiver1975:
@LukasReschke @nickvergessen please review once more - THX

nickvergessen:
sv.js language file still contains rework: https://github.com/owncloud/contacts/blob/rename-the-app/l10n/sv.js#L2

And the readme

DeepDiver1975:
@nickvergessen addressed - thx

LukasReschke:
:+1:

PR #68, opened 2016-02-29
Adding default value to field select to let user know what to do: "Ad…

DeepDiver1975:
…d field ..." :wink:


irgendwie:
LGTM :+1:

PR #69, opened 2016-02-29
Adding addressbook select on each contact

DeepDiver1975:
@Henni @jancborchardt @irgendwie @nickvergessen @LukasReschke

PR #76, opened 2016-03-01
Revert "Adding addressbook select on each contact"

Henni:
Reverts owncloud/contacts#69, which broke contact creation

cc @irgendwie


irgendwie:
:+1: let's revert that commit
didn't experience any problems while reviewing - only after a fresh setup, sorry!

PR #78, opened 2016-03-02
Always test the stable branches

nickvergessen:
https://github.com/owncloud/contacts/pull/78/files?w=1


DeepDiver1975:
I guess we will not have stable branches - but does not hurt either :+1:

PR #81, opened 2016-03-02
Set focus on added field

DeepDiver1975:


Henni:
:+1:

The focus directive looks a bit like black magic, but hey it's javascript ;)

PR #82, opened 2016-03-02
Allow multiple fields to be added

### LINKED ISSUE #73: Adding new field which already exists results in existing field being emptied
Instead of the field being emptied, there are 2 cases:
- If the field can only exist once (like »Organization« or »Nickname« or »Notes«) the entry should be removed from the »Add new field« dropdown as soon as the field already exists
- If there can be multiple fields (like telephone, address, email), there should of course be another one added.

Not sure how much of this is possible til release, but for sure we shouldn’t clear any data. As a quick fix, only focus the already existing field.

DeepDiver1975:
fixes #73 

Multiple fields can be added - this was not possible before


Henni:
@DeepDiver1975 what does this PR change/fix?

irgendwie:
LGTM :+1:

PR #92, opened 2016-03-02
Filter fields which are only allowed to be present once 

### LINKED ISSUE #89: Creation of multiple fields for Organization and Nickname possible
This should not be possible, right? Fields like organization and nickname should only exist once I assume?

DeepDiver1975:
fixes #89

### DIFF
--- a/js/filters/field_filter.js
+++ b/js/filters/field_filter.js
@@ -1,0 +1,26 @@
+app.filter('fieldFilter', [
+	function() {
+		'use strict';
+		return function (fields, contact) {
+			if (typeof fields === "undefined") {
+				return fields;
+			}
+			if (typeof contact === "undefined") {
+				return fields;
+			}
+			var filter = [];
+			if (fields.length > 0) {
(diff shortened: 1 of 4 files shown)


Henni:
Looks good so far, but can we maybe reduce the size of the `vCardMeta` object by defining a default value for `multiple`?
I believe most attributes can only be set once in vCards. Thus I would suggest to default to false and only define `multiple: true` in special cases.

DeepDiver1975:
> Looks good so far, but can we maybe reduce the size of the vCardMeta object by defining a default value for multiple?
> I believe most attributes can only be set once in vCards. Thus I would suggest to default to false and only define multiple: true in special cases.

feel free to adjust the code in this pr :wink:

Henni:
Small issue I found: adding a field which is only allowed once, leads to a blank select.
Instead it should again show "Add field".

DeepDiver1975:
rebased

DeepDiver1975:
:+1:

PR #101, opened 2016-03-03
url -> website

### LINKED ISSUE #97: Rename »URL« to »Website«
URL is much too technical, it should be called »Website«.

DeepDiver1975:
fixes #97

### DIFF
--- a/js/services/vCardProperties.js
+++ b/js/services/vCardProperties.js
@@ -24,7 +24,7 @@ app.service('vCardPropertiesService', [function() {
 		},
 		url: {
 			multiple: true,
-			readableName: t('contacts', 'Url'),
+			readableName: t('contacts', 'Website'),
 			template: 'url'
 		},
 		title: {
--- a/js/services/vCardProperties.js
+++ b/js/services/vCardProperties.js
@@ -47,7 +47,7 @@ app.service('vCardPropertiesService', [function() {
 			]
 		},
 		categories: {
-			readableName: t('contacts', 'Categories'),
+			readableName: t('contacts', 'Groups'),
 			template: 'text'
 		},
 		bday: {
--- a/js/services/vCardProperties.js
+++ b/js/services/vCardProperties.js
@@ -15,11 +15,11 @@ app.service('vCardPropertiesService', [function() {
 			template: 'text'
 		},
 		org: {
-			readableName: t('contacts', 'Organisation'),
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
@jancborchardt

jancborchardt:
Ah, that’s where it’s done! Will add some more changes on top, just a sec.

jancborchardt:
Done. Just some questions:
- What’s »Role«? And what’s the difference to »Title«? Title usually is tied to the organization, right?
- Why 3 kinds of Faxes? Can’t we cut one of them? Either the generic FAX one or the FAX, HOME one? They seem to be the same essentially?
- What’s »Categories«, and how is it used?

DeepDiver1975:
> What’s »Role«? And what’s the difference to »Title«?

https://tools.ietf.org/html/rfc6350#section-6.6.1
https://tools.ietf.org/html/rfc6350#section-6.6.2

DeepDiver1975:
> What’s »Categories«, and how is it used?

https://tools.ietf.org/html/rfc6350#section-6.7.1 - this is where we extract our groups from

DeepDiver1975:
> Why 3 kinds of Faxes? Can’t we cut one of them? Either the generic FAX one or the FAX, HOME one? They seem to be the same essentially?

these are the types I have when generating a contact using evolution - I'm not happy either.

jancborchardt:
@DeepDiver1975 ok. Can we remove »Role« then from our select? It’s really quite confusing.

DeepDiver1975:
> Can we remove »Role« then from our select? It’s really quite confusing.

and not display data which was entered on another client? not really an option from my pov

jancborchardt:
@DeepDiver1975 we can display it when it’s there, but it is confusing when we have it in our own dropdown. Where is »Role« used in Google Contacts, iOS or Android?

Also – can we then rename »Categories« to groups? Otherwise the wording will be inconsistent.

DeepDiver1975:
> Also – can we then rename »Categories« to groups? Otherwise the wording will be inconsistent.

sure

DeepDiver1975:
> we can display it when it’s there, but it is confusing when we have it in our own dropdown.

one day maybe :wink:

jancborchardt:
Hehe, ok.

DeepDiver1975:
@jancborchardt  merge?

Henni:
:+1:

Issue #98, opened 2016-03-03
After adding a field, the »Add field« select should switch back to »Add field«


jancborchardt:
Currently the select label stays at the field which was last added, which is a bit confusing. Should switch back to the default »Add field« state.

DeepDiver1975:
hmm - doesn't work any more ...

PR #102, opened 2016-03-03
More fixes for the details design

jancborchardt:


templates/detailItems/adr.html
```diff
@@ -2,10 +2,6 @@
 	<option ng-repeat="option in ctrl.availableOptions" value="{{option.id}}">{{option.name}}</option>
 </select>
 <div>
-	<label for="details-pobox">{{ctrl.t.poBox}}</label>
```
DeepDiver1975, line 5: @jancborchardt please leave the po box in here - it's a required part of the address

https://tools.ietf.org/html/rfc6350#section-6.3.1

DeepDiver1975, line 5: there is even one more we need to add - the second ' the extended address (e.g., apartment or suite number);'

jancborchardt, line 5: Ugh, way to make a spec complicated. :\ P.O. box won’t be used in the majority of cases anyway, similar to extended address.

DeepDiver1975:
30db794 fixes #98

jancborchardt:
kk, let’s merge?

jancborchardt:
Ah ok, Travis is required here as well, sorry. :)

DeepDiver1975:
> Ah ok, Travis is required here as well, sorry. :)

travis is setup to execute phpunit tests - we have no php code in here ...

jancborchardt:
Still merging was seemingly blocked for me.

DeepDiver1975:
> Still merging was seemingly blocked for me.

sure .... you are not an admin :wink:

DeepDiver1975:
we need to sort out the ci situation in here ... next meetup maybe ...

Issue #100, opened 2016-03-03
Default select state (Home, Work, etc) when adding new address or phone number


jancborchardt:
Currently it’s an empty select, which looks really weird. And can it even be empty?

It should be the first entry (Private) by default. If one already exists with private, use the next one (Work). And so on. This is how Android does it as well.

PR #104, opened 2016-03-03
- Pre-select address type on adding a new address

DeepDiver1975:
- Fix reset of field adding select


DeepDiver1975:
fixes #100
fixes #98

tomneedham:
:+1: Fixes both those issues

Issue #99, opened 2016-03-03
Also add select for Email, Website and IM?


jancborchardt:
They should also have the Private / Work etc dropdown, no?

DeepDiver1975:
im : https://tools.ietf.org/html/rfc6350#section-6.4.3
email: https://tools.ietf.org/html/rfc6350#section-6.4.2
url: should have as well

Henni:
fixed by #107

PR #107, opened 2016-03-03
adding types selects to email, impp and tel

DeepDiver1975:
refs #99

### DIFF
--- a/js/public/script.js
+++ b/js/public/script.js
@@ -1154,17 +1154,39 @@ app.service('vCardPropertiesService', [function() {
 		email: {
 			multiple: true,
 			readableName: t('contacts', 'Email'),
-			template: 'text'
+			template: 'text',
+			defaultValue: {
+				value:[''],
+				meta:{type:['HOME']}
+			},
+			options: [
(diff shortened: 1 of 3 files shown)


js/public/script.js
```diff
@@ -1154,17 +1154,39 @@ app.service('vCardPropertiesService', [function() {
...
+			template: 'text',
+			defaultValue: {
+				value:[''],
+				meta:{type:['HOME']}
+			},
+			options: [
+				{id: 'HOME', name: t('contacts', 'Home')},
+				{id: 'WORK', name: t('contacts', 'Work')},
+				{id: 'OTHER', name: t('contacts', 'Other')}
+			]
```
irgendwie, line 1180: Instant messaging should be separated by the protocol (AIM, Jabber, MSN, etc.) and not by categories (Home, Work, etc.)!

DeepDiver1975, line 1180: the protocol is part of the value - https://tools.ietf.org/html/rfc6350#section-6.4.3

irgendwie, line 1180: whoops, my fault - sry!

irgendwie:
LGTM :+1:

PR #108, opened 2016-03-03
add federated cloud id - fixes #106

### LINKED ISSUE #106: Allow user to add federated cloud id to contact
I just tried the new contacts app and first let me say that it looks already really good! :-)

The old contacts app allowed the user to add th federated cloud ID to a contact.

The cloud id was stored this way in the vcard:

```
CLOUD;TYPE=HOME:foo@cloud.owncloud.org
```

It supported the same types as the email field so that a user could have multiple federated cloud IDs, e.g one personal ID and one work ID.

The share dialog also searches the address book for users with a cloud id, so this is quite useful to share with people from your address book. Would be great if this field would also exist in the new contacts app.

DeepDiver1975:


irgendwie:
LGTM :+1:

schiessle:
This was fast :)

Looks good, but I can't specify a type like "work", "home", etc. Did you skipped it by intention? Or is this a current limitation of the app, because I see that the field for the email address also don't support this types.

This also causes problems for the auto-complete. Because if you add multiple federated cloud IDs, the auto complete in the share dialog will always return the last one.

EDIT: I just looked at the code and saw that there you specified different types. So maybe it is just a UI issue that I can't select them?

schiessle:
I will check why the search only return the latest cloud id, maybe you can check why the UI doesn't show the types and let me chose one.

js/public/script.js
```diff
@@ -1129,6 +1129,19 @@ app.service('vCardPropertiesService', [function() {
 			readableName: t('contacts', 'Role'),
 			template: 'text'
 		},
+		cloud: {
+			multiple: true,
+			readableName: t('contacts', 'Federated Cloud Id'),
```
jancborchardt: ID should have both letters capitalized. Same for the vCardproperties file.

jancborchardt: Btw … why is the code basically duplicated?

DeepDiver1975:
@schiesbn  this branch needs a rebase after merge of #107

js/public/script.js
DeepDiver1975: Only script.js is loaded by the Browser. This file is concatenated from all the others.

DeepDiver1975:
bildschirmfoto von 2016-03-04 08-58-12

@jancborchardt @schiesbn this is how it looks like now

schiessle:
looks good now :+1:

PR #109, opened 2016-03-03
Default to home type for tel number

tomneedham:
Pre-selects 'Home, Voice' as the phone number type just like for the address.


DeepDiver1975:
already done here - https://github.com/owncloud/contacts/pull/107/files#diff-dbaa024398385bf6cffe22c422471ff0R1186

tomneedham:
We're all too quick ;) At least it was a 20 second job.

PR #120, opened 2016-03-04
Fix birthday field 

### LINKED ISSUE #112: Using datepicker causes crash
Use the datepicker, as soon as you pick a date the JS console spits out something like this:

```
TypeError: value.value.map is not a function
```

(Long error, best test for yourself)

And after that, no fields can be added or deleted until refresh.

DeepDiver1975:
fixes #112

@jancborchardt


Henni:
:+1:

Henni:
text-based birthdays in #122

DeepDiver1975:
> text-based birthdays in #122

thx

PR #121, opened 2016-03-04
Fix some field processing

### LINKED ISSUE #113: Only show url-link arrow element for website when an actual value was put in
Otherwise it makes no sense to show the linking arrow as it has no function.

### LINKED ISSUE #117: When adding a new contact, insert some detailItems by default
Currently there’s only an empty view with the name input and you need to add fields by yourself.

Instead, these common fields should be pre-inserted, in that order:
- Phone
- Email
- Address

DeepDiver1975:
fixes #117 
fixes #113

### DIFF
--- a/templates/detailItems/url.html
+++ b/templates/detailItems/url.html
@@ -6,4 +6,4 @@
 	   focus-expression="$parent.ctrl.focus === ctrl.name"
 />
 <button ng-click="ctrl.deleteField()" class="icon-delete" title="Delete"></button>
-<a class="url-link" target="_blank" href="{{ctrl.data.value}}">↗</a>
+<a ng-if="ctrl.data.value.length" class="url-link" target="_blank" href="{{ctrl.data.value}}">↗</a>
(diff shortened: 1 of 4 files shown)


templates/detailItems/url.html
```diff
@@ -6,4 +6,4 @@
 	   focus-expression="$parent.ctrl.focus === ctrl.name"
 />
 <button ng-click="ctrl.deleteField()" class="icon-delete" title="Delete"></button>
-<a class="url-link" target="_blank" href="{{ctrl.data.value}}">↗</a>
+<a ng-if="ctrl.data.value.length" class="url-link" target="_blank" href="{{ctrl.data.value}}">↗</a>
```
Henni, line 9: Angular automatically sets `ng-invalid` and `ng-invalid-url` classes. We could use them and only show the link if the url is valid. This could be done in CSS.

BTW: we should use ng-href

Henni, line 9: I experimented a bit on this. Seems like the pattern angular uses requires `http://` or `https://`.

Henni:
LGTM :+1:

jancborchardt:
CONTACTS TEAM ON FIRE :rocket:

Issue #119, opened 2016-03-04
Favorites


jancborchardt:
Will favorites work for 9.0? Will it be a read-only indicator maybe? Or not at all?

Either way, if we don’t finish that, we should hide the star in the contact list and then move the avatar a tad down for visual balance.

DeepDiver1975:
> we should hide the star in the contact list and then move the avatar a tad down for visual balance.

let's do that - we need to finish 1.0 by Monday

PR #123, opened 2016-03-04
Hide star

jancborchardt:
Ref #119 

Also I made a style so that it’s nicer when the contact has no second line yet. @Henni can you make it so that the element `.app-content-list-item-line-one` gets the class `.no-line-two` when there’s no content in `.app-content-list-item-line-two`?

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -181,12 +181,12 @@ detailsitem .icon-delete {
 	position: absolute;
 	display: inline-block;
 	left: 10px;
-	top: 5px;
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
:+1:

jancborchardt:
@Henni @irgendwie just as an enhancement and only if someone of you has time :)

> Also I made a style so that it’s nicer when the contact has no second line yet. @Henni can you make it so that the element `.app-content-list-item-line-one` gets the class `.no-line-two` when there’s no content in `.app-content-list-item-line-two`?

PR #127, opened 2016-03-04
display email again in the contacts list 

### LINKED ISSUE #124: Second line not working anymore
Previously the second line was automatically populated with a value like the email address. Now this doesn’t happen anymore it seems.

It should show the organization, or if there’s none set for that contact, the phone number – if that’s not set either, the email address.

If neither are set, then show nothing, also as some kind of a hint about that contact having incomplete info. (We could also fall back further but I see no real use in that.)

DeepDiver1975:
fixes #124 

@jancborchardt

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -32,6 +32,16 @@ app.factory('Contact', [ '$filter', function($filter) {
 				}
 			},
 
+			email: function() {
+				// getter
+				var property = this.getProperty('email');
+				if(property) {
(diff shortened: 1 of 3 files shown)


irgendwie:
If no e-mail address is present (for example a new contact is created),

> [""]

is shown instead of an empty space.

DeepDiver1975:
@irgendwie fixed

irgendwie:
Nice! :+1:

PR #129, opened 2016-03-04
Show loading spinners whilst retriving data

tomneedham:
- [ ] Need to vertically align the spinners

### DIFF
--- a/js/services/addressBook_service.js
+++ b/js/services/addressBook_service.js
@@ -1,6 +1,7 @@
 app.factory('AddressBookService', function(DavClient, DavService, SettingsService, AddressBook, Contact) {
 
 	var addressBooks = [];
+	var loading = true;
 
 	var loadAll = function() {
 		return DavService.then(function(account) {
--- a/js/services/addressBook_service.js
+++ b/js/services/addressBook_service.js
@@ -1,7 +1,6 @@
 app.factory('AddressBookService', function(DavClient, DavService, SettingsService, AddressBook, Contact) {
 
 	var addressBooks = [];
-	var loading = true;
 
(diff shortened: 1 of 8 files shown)


jancborchardt:
Just use a div to cover the whole area and give it the class `.loading` – that’s how we do it everywhere else as well. ;)

tomneedham:
ok

jancborchardt:
Updates here?

tomneedham:
So as you say the best solution is to add `.loading` to here: https://github.com/owncloud/contacts/blob/master/templates/main.php#L44 

But my Angular knowledge runs out here, @Henni is this even possible as its not in a template?

Henni:
You could probably apply it to the contactlist instead, which has its own controller.
https://github.com/owncloud/contacts/blob/master/templates/main.php#L45

tomneedham:
Ready for review @jancborchardt

DeepDiver1975:
I see a spinner! :tada:

DeepDiver1975:
needs rebase

DeepDiver1975:
:+1:

js/services/addressBook_service.js
```diff
@@ -1,6 +1,7 @@
 app.factory('AddressBookService', function(DavClient, DavService, SettingsService, AddressBook, Contact) {
 
 	var addressBooks = [];
+	var loading = true;
```
irgendwie: Do we need this variable here?

irgendwie:
Rebased!
Pls review @Henni @DeepDiver1975

Henni:
LGTM :+1:

PR #131, opened 2016-03-04
Order detail items

### LINKED ISSUE #118: Order of detail items
Currently the fields are ordered alphabetically in the dropdown, and by i-have-no-idea in the details view.

The order should be like this in both the details view and dropdown (**bold** items are shown by default):
- Organization & title (should be combined

^ we could put that in the header as well, smaller below the name
- **Phone**
- **Email**
- **Address**
- Instant messaging

^ that’s all the contact stuff, which can have home/work/other dropdowns.
- Nickname
- Birthday
- Website
- Notes
- Groups
- Role (let’s remove it …)

^ not entirely set on the order of the stuff in this last part.

Henni:
fixes #118 

I put the order of the fields into a separate array, because objects aren't guaranteed to be ordered.

@DeepDiver1975 please review
@jancborchardt I didn't order the select. Do we really want this? I'd prefer it sorted alphabetically.

### DIFF
--- a/js/filters/orderDetailItems_filter.js
+++ b/js/filters/orderDetailItems_filter.js
@@ -1,0 +1,27 @@
+app.filter('orderDetailItems', ['vCardPropertiesService', function(vCardPropertiesService) {
+	'use strict';
+	return function(items, field, reverse) {
+
+		var filtered = [];
+		angular.forEach(items, function(item) {
+			filtered.push(item);
+		});
+
+		var fieldOrder = angular.copy(vCardPropertiesService.fieldOrder);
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
> I didn't order the select. Do we really want this? I'd prefer it sorted alphabetically.

I'd keep the select in alphabetic order as well

DeepDiver1975:
:+1:

DeepDiver1975:
rebased

jancborchardt:
@Henni I would prefer the select to be ordered the same way because it creates a symmetry of sorting, and fulfills expectations.

PR #134, opened 2016-03-04
Fix classes of detail items

Henni:
fixes #133
@jancborchardt

PR #135, opened 2016-03-04
Allow invalid urls

Henni:
fixes that urls aren't saved, if they don't match angular's url pattern.


irgendwie:
Does **not** work for me :-1:

Henni:
@irgendwie this might have been because #134 was missing.
Rebased and should work now

irgendwie:
Seems so, works now as expected. :+1:

Still, if you add an URL not starting with http,
in the clickable arrow it is treated (now) as a relative path beginning from `index.php/apps/`.

As example:

> URL: asdf

leads to

> http://localhost/oc/index.php/apps/asdf

Henni:
@irgendwie that's a different issue. (and probably not in scope for 1.0)

irgendwie:
@Henni just wanted to address this issue, since it will be introduced with this PR.

PR #136, opened 2016-03-04
Fix typeError when no contact is available.

irgendwie:
`Uncaught (in promise) TypeError: Reduce of empty array with no initial value(…)`

To reproduce: Start contacts app without any contact available.

This PR adds an initial value for the reduce function to overcome this error.


Henni:
LGTM :+1:

Henni:
rebased

PR #138, opened 2016-03-04
Design details

jancborchardt:
Mostly settings area bottom left, please review @Henni @irgendwie (no work on the sharing list yet)


tomneedham:
screen shot 2016-03-04 at 17 15 18

Looks like the icons are slightly raised from the addressbook names, but :+1:

tomneedham:
Also: 
screen shot 2016-03-04 at 17 16 51
but guessing this is a separate issue not caused by this.

PR #141, opened 2016-03-05
Fix #132.

### LINKED ISSUE #132: Align first line in row nicely if there’s no content in second line
When there’s no content in `.app-content-list-item-line-two`, then `.app-content-list-item-line-one` should get the additional class `.no-line-two`. That will make it nicely layouted.

(Creating an issue so we don’t lose track.)

irgendwie:
@jancborchardt

### DIFF
--- a/templates/contact.html
+++ b/templates/contact.html
@@ -4,6 +4,6 @@
 
 <div class="app-content-list-item-star icon-star" data-starred="false"></div>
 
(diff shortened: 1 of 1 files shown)


irgendwie:
Dynamic updates in this template are currently broken (also on pure master) until you create a new contact.
Any ideas on that @Henni?

Henni:
LGTM :+1: 

https://github.com/owncloud/contacts/pull/141#issuecomment-192539822 is a strange bug. I'll look into it.

PR #146, opened 2016-03-05
Keep database structure of old contacts app

Henni:
Makes it possible to migrate even after upgrading the contacts app.

cc @BernhardPosselt @DeepDiver1975


Henni:
False alarm

PR #157, opened 2016-03-07
remove unnecessary background for non-existent profile image

jancborchardt:
It’s not necessary and just alludes at the possibility that photo upload is possible. Which it isn’t yet.

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -8,8 +8,9 @@
 }
 
 .contactdetails__header h2 {
-	margin-bottom: 6px;
-	margin-left: 8px;
+	position: absolute;
+	top: 25px;
+	left: 103px;
 }
 .contactdetails__header .contactdetails__name {
 	font-size: inherit !important;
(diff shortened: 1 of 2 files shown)


tomneedham:
With this PR:
screen shot 2016-03-07 at 15 20 10
without: 
screen shot 2016-03-07 at 15 20 46
... in Safari on OSX

tomneedham:
Trying to work out where that border is coming from

tomneedham:
... Not sure. Works in Firefox but Chrome/Safari I seem to see a 1px border on the `img` tag but it doesnt actually show in the inspector

DeepDiver1975:
Should we simply hide the logo element in case there is no photo?

@jancborchardt

jancborchardt:
@tomneedham yeah, I was confused too. No idea where it’s coming from.

@DeepDiver1975 right, probably best choice. I added a commit to absolutely position the name so layout isn’t broken when the img is hidden. You can go for it and implement it in this PR.

jancborchardt:
(And rebased)

DeepDiver1975:
> Should we simply hide the logo element in case there is no photo?

done

jancborchardt:
Thanks!

PR #158, opened 2016-03-07
Use jquery datepicker for BDAY

### LINKED ISSUE #143: Birthdays don't get saved
Hi,

I am using the last master and try to add some sample contacts at the moment I have just issues to get the birthdays in. If I add the field Birthday and hover over it the mouse changes to an hand but nothing else happens even if I click into the field. If I write a birthday manually in like 01.01.1999 it sticks for the moment but disappears as son the page get refreshed.
What should happen as soon I am hovering with the mouse over the field? Would I see an input calendar?
Thanks in advance

DeepDiver1975:
fixes https://github.com/owncloud/contacts/issues/143


Henni:
looks good and owncloud branded :+1:

MeCias:
Works perfectly with Firefox 45. Thanks a lot.

Issue #49, opened 2016-02-19
Integrate Search 


aspdye:

aspdye:
Are we releasing this with 2.0 or can we integrate it with 1.1 :question:

DeepDiver1975:
will be in 1.1

PR #164, opened 2016-03-07
Added search functionality.

irgendwie:
Adds #49.
@jancborchardt


jancborchardt:
NICE! Already works pretty well. Just some things, sorted by most important first:
- [x] When there is no result, there should be an empty state, with the icon-search and a message below – just copy that from the Mail app.
- [x] Changing the group using the navigation on the left should reset the search.
- [x] Can pressing enter select that result, that is open the first contact?
- [x] Currently only the name is searched. Can that be expanded to other fields like address, and phone number?

jancborchardt:
(I rebased btw)

Henni:
Still stuff to do, but looks good so far.

Henni:
???
image

irgendwie:
Fix for last bullet point in #218.

aspdye:
According to #49, i think we should change the milestone to 1.1 :wink: #justfordocumentation

PR #165, opened 2016-03-08
Always show groups field and don't allow to delete this field 

### LINKED ISSUE #163: Add groups field to default fields of new contacts
The groups field should be a default field for new contacts, and all contacts.

It also doesn’t make a whole lot of sense to be able to remove this, so the deletion functionality should be disabled.
Basically same for the last email, phone, and address – these should always stay there by default. Just as you can not remove the name.

DeepDiver1975:
fixes #163

@jancborchardt


jancborchardt:
Nice! :+1:

PR #168, opened 2016-03-08
Org and title are now part of the header

### LINKED ISSUE #103: Make Organization and Title a group
Similar to the address fields, the fields for »Organization« and »Title« belong together and should be made a group. Title should always appear below the Organization field. That’s also how Android and others do it.

Visually, we could then also add some whitespace above and below that group similarly to how we do with the address.

DeepDiver1975:
fixes #103 @jancborchardt @tomneedham 

bildschirmfoto von 2016-03-08 12-05-01

Only placeholders if not set:
bildschirmfoto von 2016-03-08 12-05-46

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -19,6 +19,24 @@
 	background: transparent !important;
 	border: none !important;
 }
+.contactdetails__header .contactdetails__org {
+	font-size: inherit !important;
+	width: 100% !important;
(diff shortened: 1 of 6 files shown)


jancborchardt:
@DeepDiver1975 fixed the styling. Nice! :+1: 

Please rebase.

PR #187, opened 2016-03-10
Add elipses to name, org, title in contact details header

tomneedham:
Make things look a little nicer in the header with long companies, titles, names etc:

Before:
screen shot 2016-03-10 at 15 11 35

After:
screen shot 2016-03-10 at 15 08 52

You can still click in the fields and edit the full value and the full value is stored.

@jancborchardt  @Henni

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -20,6 +20,9 @@
 	color: #fff !important;
 	background: transparent !important;
 	border: none !important;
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
I like it :+1:

Curatio:
Very nice.

Issue #193, opened 2016-03-11
Contact photos are always displayed with square aspect ratio, and without crop


nicokaiser:
### Steps to reproduce
1. Create a contact with Mac OS X Contacts app
2. Add a non-square picture and crop it, save the contact
3. Open the contact in ownCloud
### Expected behaviour

The picture should be displayed correctly cropped
### Actual behaviour

Seems like OS X saves the un-cropped image along with cropping information, which ownCloud cannot apply. So the images is displayed uncropped yet with a square aspect ratio, which may both be problematic. See screenshot.

<img width="430" alt="screen shot 2016-03-11 at 15 03 31" src="https://cloud.githubusercontent.com/assets/238631/13704201/f34cfd4e-e79a-11e5-85ee-2c4930f5ffae.png">
### Server configuration

**Operating system**: Debian 8

**Web server:** nginx

**Database:** MySQL 5.5

**PHP version:** PHP 5.6

**ownCloud version:** 9.0.0

**Contacts version:** 1.0.0.0

**Updated from an older ownCloud or fresh install:** Updated from 8.x

**Signing status (ownCloud 9.0 and above):**

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
  - activity: 2.2.1
  - calendar: true
  - comments: 0.2
  - contacts: true
  - dav: 0.1.5
  - federatedfilesharing: 0.1.0
  - federation: 0.0.4
  - files: 1.4.4
  - files_sharing: 0.9.1
  - files_texteditor: 2.1
  - files_trashbin: 0.8.0
  - files_versions: 1.2.0
  - files_videoplayer: 0.9.8
  - firstrunwizard: 1.1
  - gallery: 14.5.0
  - news: true
  - notifications: 0.2.3
  - provisioning_api: 0.4.1
  - systemtags: 0.2
  - templateeditor: 0.1
  - updatenotification: 0.1.0
Disabled:
  - encryption
  - external
  - files_external
  - files_pdfviewer
  - tasks
  - user_external
  - user_ldap
```

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "512fdebce061d",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "datadirectory": "\/var\/www\/cloud....\/data",
        "dbtype": "mysql",
        "version": "9.0.0.19",
        "dbname": "owncloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
[...]

PR #196, opened 2016-03-11
Keep the aspect of the contact images

DeepDiver1975:
refs #193 

It does not yet evaluate the crop information as specified in the vcard field - but at least the aspect is kept

@nicokaiser 

bildschirmfoto von 2016-03-12 00-39-57

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -78,6 +78,7 @@
 .contactdetails__logo {
 	height: 75px;
 	width: 75px;
+	object-fit: cover;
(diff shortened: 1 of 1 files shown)


Henni:
Doesn't work on IE or Edge (http://caniuse.com/#feat=object-fit), otherwise looks good.

DeepDiver1975:
> Doesn't work on IE or Edge

We can work on a IE fix in a separate PR

PR #218, opened 2016-03-17
Fix no search result message.

irgendwie:
@Henni My thought on the train was correct! :blush:
(And you can do the CSS stuff now :smile:)


Henni:
rebased and :+1:

PR #226, opened 2016-03-17
Remove IRC info from README.md

aspdye:
#223 added an IRC badge, so the notice at the end of the README.md is not longer needed :wink:

cc @irgendwie @DeepDiver1975


Henni:
@irgendwie and I talked about this a few minutes ago. We thought it doesn't hurt to have it twice. But I'm also happy with cleaning it up and removing the occurrence at the end.

@aspdye Feel free to merge, if you like it better this way.

PR #236, opened 2016-03-19
Remove globals and warn for unused variables

Henni:
The eslint output is a bit spammy at the moment because there are a lot of unused variables in our code.


irgendwie:
@Henni How to review this? Run eslint?

PR #237, opened 2016-03-19
vCard Import

Henni:
Basic import functionality.
Currently the import functionality is in the settings on the bottom left.
It still has to be restyled and probably moved.

### DIFF
--- a/js/components/contactImport/contactImport_controller.js
+++ b/js/components/contactImport/contactImport_controller.js
@@ -1,0 +1,7 @@
+angular.module('contactsApp')
+.controller('contactimportCtrl', function(ContactService) {
+	var ctrl = this;
+
+	ctrl.import = ContactService.import.bind(ContactService);
+
+});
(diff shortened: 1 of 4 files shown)


tomneedham:
Drag and drop import? ;)

Henni:
@tomneedham sure. In the future ;)
In my opinion we should offer a button to make this feature discoverable, and _additionally_ allow drag and drop.

tomneedham:
Sounds good - will test in a few minutes

tomneedham:
I imported a vcard with about 20 contacts in and it resulted in one contact being created with all of the contacts details inside the name of the first contact in the file.

irgendwie:
A single contact works as expected, but not a *.vcf with many contact in it.
Maybe we should iterate over `BEGIN:VCARD` and `END:VCARD` in the imported file.

Henni:
A simple split over `END:VCARD` might fix this.
@irgendwie do you want to take over?

irgendwie:
@Henni I think this issue isn't that easy :disappointed: 
The old contacts app had different formats for the import function:
https://github.com/owncloudarchive/contacts/tree/master/formats

Maybe we should think of a way to use them!

Henni:
@irgendwie new commit

irgendwie:
Nice work! :+1: VCard import should be enough for the next version.
For the styling, why not do it like the old contacts app?

settingspane

PR #241, opened 2016-03-20
Fixed spanish translation.

ghost:
"Home" being translated as "Start" was very confusing.


Henni:
I believe this should be done via transifex.
@DeepDiver1975 can you verify this?

DeepDiver1975:
Yes we need to fix this on transifex

PR #261, opened 2016-03-22
Better colors generator

skjnldsv:
A better color generator, with pastel colors and a brightness check to prevent illisible text.

I didn't liked the colors on the contacts app, so i made a color generator using the hsl wheel and a different angle base on the contact uid. I use a hsl to rgb function found online (too simple to bother coding the same function on my own) and a basic common brightness algorithm.

capture d ecran_2016-03-22_09-26-21


irgendwie:
@skjnldsv Would you mind squashing your last commit? (https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)
LGTM :+1:

skjnldsv:
@irgendwie Good here?

jancborchardt:
@skjnldsv can you directly modify the color generator in core please? See https://github.com/owncloud/core/blob/master/core/js/placeholder.js#L54-L58 :)

Because we should use the same color scheme across ownCloud, especially for the placeholders. When I share something with a person, the placeholder should be the same as in the Contacts app for that person.

skjnldsv:
Dammit, why no one said that before? :p

jancborchardt:
@skjnldsv yeah, sorry bout that. ;) Would still be cool if you could do that. :)

skjnldsv:
Too busy crying ;)
Will do in a moment, i'm working on the header inputs now.

jancborchardt:
:D Welcome to open source development ;D

skjnldsv:
Wait a minute, I didn't realised. You want to use the core generator in the contacts app, or my generator in core?
If you were considering the second option, please see #306

jancborchardt:
@skjnldsv a combination of both: Ideally your improvements are done to the core generator, and then we use the core generator in Contacts. (Because the core generator is also used everywhere else.)

PR #280, opened 2016-03-25
Fix setSelected not correcty called in groupList.html.

irgendwie:
**How to test:**
- Search for something
- Click on a different group

**Expected result:**
New group is selected and search is cleared.

**Actual result:**
New group is selected but search isn't cleared.


DeepDiver1975:
tested works :+1:

PR #282, opened 2016-03-25
Integrate codecov.io

DeepDiver1975:


Henni:
A gitignore entry for the coverage data wouldn't hurt.
Otherwise :+1:

DeepDiver1975:
@Henni tests are not yet running on travis and fail because chrome cannot be started .....

ideas are welcome :wink:

DeepDiver1975:
https://karma-runner.github.io/0.8/plus/Travis-CI.html

Henni:
@DeepDiver1975 we will be able to switch to PhantomJS after @irgendwie removed the dav lib.
The dav lib relies on ES6's `Set` which isn't supported by PhantomJS (https://kangax.github.io/compat-table/es6/#phantom)

Henni:
A bit of research tells me that Firefox is preinstalled on travis machines. We could test in Firefox instead of Chrome.

But let's stick with Chrome if something like this works:

```
before_script:
    - 'export CHROME_BIN=chromium-browser'
    - 'export DISPLAY=:99.0'
    - 'sh -e /etc/init.d/xvfb start'
```

Henni:
experimenting at #285

Henni:
The [Chrome Version on Travis](https://travis-ci.org/owncloud/contacts/builds/118483071#L472) is 37!!! This version is from 2014-08-26!
Additionally the travis logs show a lot of warnings because it uses the node version 0.10 (latest stable: 5.9.1)

We might have to use a different travis setup and install node and chrome ourselves.

DeepDiver1975:
Lets pull in underscore with Bower. Otherwise we would need to install a full ownCloud core

Henni:
fixed :)

DeepDiver1975:
Nice

PR #287, opened 2016-03-25
Travis firefox

Henni:


Henni:
@DeepDiver1975 except for underscore.js this works!

DeepDiver1975:
Nice. Is underscore loaded in the Karma config?
Worstcase we installl it for testing using bowser

Henni:
Mario wouldn't be happy if we install it with bowser ;)
Currently we use the version provided by core, which isn't loaded in travis. We could add it to the devDependencies of bower instead.

I'll close this and add a commit to the codecov.io PR.

PR #290, opened 2016-03-25
Fix #279 - import button too big in some languages

### LINKED ISSUE #279: Import button overlaps settings button in german ui
In german ui the import button overlaps the settings button.

**ownCloud version:** 9.0.0
**Contacts version:** 1.1.0

skjnldsv:
Fix #279, overlapping of the import button in some languages.

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -285,15 +285,15 @@ li.addressBook-share-item span.shareeIdentifier {
 }
 
 #app-settings-content #upload.button {
-	padding: 5px 40px 5px 20px;
-	background-position-x: 80px;
+	padding: 5px;
+	padding-right: 25px;
 	position: absolute;
 	top: 2px;
 	right: 2px;
 }
 #app-settings-content #upload.button .icon-upload {
 	display: block;
-	right: 16px;
+	right: 5px;
 	position: absolute;
 	top: 7px;
 	opacity: 0.5;


Henni:
@skjnldsv as you are a member of owncloud, feel free to create branches directly inside this repository.
This simplifies the review process and working together on PRs.

skjnldsv:
@Henni, Thanks, but I don't want to risk and make a mess :p 
Do be faire, i'm not the greatest git user and better safe than sorry ^^

skjnldsv:
But I will do next time, it shouldn't be that different that creating a fork on my profile! ;)

Henni:
Better but not perfect. Maybe we should find a solution that works independent of text width, in case of languages with even longer words than German.

@owncloud/designers what do you think?
Before:
image
now:
image

skjnldsv:
Well, there's multiple solution:
1. We set the index of the "Settings" text to always have it above the import button
2. We remove the "Settings text" when the setting section is open
3. We set a max-width on the import button.

What do you prefer?
(alternatively, we can also just move the import button, but I liked the location of it ;) )

jancborchardt:
@skjnldsv I gotta say I like the idea of placing the import button there. However with the translation issue it’s a bit of a problem.

I would say moving the import button below the settings header in its own line might be best. Giving it full width, might also make the settings look a bit simpler.

skjnldsv:
Closing, we need to move it into a specific menu bar inside the settings.
Waiting for more option to add (see #297)

PR #293, opened 2016-03-25
Org and title on separate lines - Fixes #286

### LINKED ISSUE #286: contacts v1.1.0.0 - company and job role information almost impossible to read
## Expected behavior

When editing and/or displaying the contact details, the company and job role information should
be shown in full length, one per line.
## Current behavior

The company and job role information is cropped and displayed in a single line which makes it
almost impossible to read it correctly. There is sufficient space on a desktop PC or tablet PC screen
and no need to crop the information.

Current layout:
20160325_contact_details_cropped

Desired layout:
20160325_contact_details_cropped-2
#### Server Configuration

OS: Linux 3.2.75
Web server: Apache2 2.4.18
Database: MariaDB 5.5.46
PHP version: 5.6.17
ownCloud version: 9.0.0
Calendar app version: 1.1.0.0
#### Client Configuration

Browser: Firefox 45.0.1
Operating system: Windows 7

tomneedham:
Before:

screen shot 2016-03-25 at 17 12 03
screen shot 2016-03-25 at 17 11 58

After:

screen shot 2016-03-25 at 17 10 33
screen shot 2016-03-25 at 17 10 23

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -9,14 +9,14 @@
 
 .contactdetails__header h2 {
 	position: absolute;
-	top: 15px;
+	top: 8px;
 	left: 100px;
 }
 .contactdetails__header .contactdetails__name,
 .contactdetails__header .contactdetails__org,
 .contactdetails__header .contactdetails__title {
 	font-size: inherit !important;
-	width: 100% !important;
+	width: 60% !important;
 	color: #fff !important;
 	background: transparent !important;
 	border: none !important;
@@ -26,13 +26,13 @@
 }
 .contactdetails__header .contactdetails__org {
 	position: absolute;
-	left: 101px;
-	width: 100px !important;
+	left: 100px;
+	top: 36px;
 }
 .contactdetails__header .contactdetails__title {
 	position: absolute;
-	left: 201px;
-	width: 125px !important;
+	left: 100px;
+	top: 58px;
 }
 /* fix placeholder color */
 .contactdetails__header .contactdetails__name::-webkit-input-placeholder,


tomneedham:
https://github.com/owncloud/contacts/issues/286

DeepDiver1975:
Nice. I like it.  @jancborchardt do you agree?

skjnldsv:
Could be interesting to space things a little to be a little more readable if you may. :)
Like this for example? Or not, like you want! ;)

ds
dsa

jancborchardt:
As said in https://github.com/owncloud/contacts/issues/286#issuecomment-201927872 I do not really like the two lines as they take too much importance.

And adding an indent would make it look even stranger, sorry @skjnldsv – everything should be flush aligned.

We can increase the width of both fields a bit, that would improve the situation.
In an ideal world where this works for all languages, we could make it even more »speaking« like

> `Lead dev` at `Supercorp`

and the first field could have a variable width so it’s nicely laid out.

tomneedham:
What about on mobile @jancborchardt? I tried various combinations of titles and orgs (using real life values) and with the avatar you can never get it to display correctly on one line.

jancborchardt:
Then not sure if it’s the best idea to put it up there at all. Because actually maybe I have that info filled out for 10% or 20% of my contacts. If it’s such a fringe use it doesn’t belong so presently, especially when we right now need to jump through hoops to make it work.

skjnldsv:
No problem @jancborchardt ! :)
Maybe this shouldn't be up there at all indeed. But for sure we need to figure out the best way to displays these info.

skjnldsv:
Or can we leave them on the same line, but set a min-width, a max-width and make the input grow from min to max? It should allow small and big value without any problem and still not take too much space.

jancborchardt:
@skjnldsv yeah, good call on the min- and max-width. Wanna whip up a pull request for that?

skjnldsv:
Could be a little difficult to do, but will try.

skjnldsv:
See here: https://github.com/owncloud/contacts/commit/32eae66969e3e824f9933590c5726eae12a1628b and here: #314

jancborchardt:
Closing in favor of https://github.com/owncloud/contacts/pull/314

PR #306, opened 2016-03-27
Improved color generator and fix #300

### LINKED ISSUE #300: Invisible contacts letter avatar
### Expected behaviour

Contact list with different colors.
### Actual behaviour

Some contacts avatars are invisible. Selecting the list make the letter appears.

capture

Also, those contact edit page looks like this:

### Server configuration

**Operating system**: Debian

**Web server:** Apache

**Database:** MySQL

**PHP version:** 5.6

**ownCloud version:** 9.0.0

**Contacts version:** 1.1.0.0

**Updated from an older ownCloud or fresh install:** From 8.2.3

**Signing status (ownCloud 9.0 and above):** Got extra files errors after upgrade, deleted each files afterward.

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
  - activity: 2.2.1
  - calendar: 1.0
  - comments: 0.2
  - contacts: true
  - dav: 0.1.5
  - documents: 0.12.0
  - federatedfilesharing: 0.1.0
  - federation: 0.0.4
  - files: 1.4.4
  - files_locking: true
  - files_pdfviewer: 0.8
  - files_sharing: 0.9.1
  - files_texteditor: 2.1
  - files_trashbin: 0.8.0
  - files_versions: 1.2.0
  - files_videoplayer: 0.9.8
  - firstrunwizard: 1.1
  - gallery: 14.5.0
  - notifications: 0.2.3
  - provisioning_api: 0.4.1
  - systemtags: 0.2
  - templateeditor: 0.1
  - updatenotification: 0.1.0
Disabled:
  - encryption
  - external
  - files_external
  - files_videoviewer
  - galleryplus
  - pdflintview
  - updater
  - user_external
  - user_ldap
  - user_webdavauth
```

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc26ab9bf921",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "domain.com"
        ],
        "datadirectory": "\/var\/www\/domain.com\/data",
        "dbtype": "mysql",
        "version": "9.0.0.19",
        "dbname": "owncloud_db",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "installed": true,
        "theme": "",
        "maintenance": false,
        "secret": "***REMOVED SENSITIVE VALUE***",
        "loglevel": 2,
        "forcessl": true,
__ MAIL INFO REMOVED__,
        "forceSSLforSubdomains": true,
        "trashbin_retention_obligation": "auto"
    }
}

```

**Are you using external storage, if yes which one:** No

**Are you using encryption:** No

**Are you using an external user-backend, if yes which one:** Nope

skjnldsv:
Better colors randomisation.

### DIFF
--- a/js/filters/contactColor_filter.js
+++ b/js/filters/contactColor_filter.js
@@ -1,41 +1,42 @@
 angular.module('contactsApp')
 .filter('contactColor', function() {
 	return function(input) {
-		function hslToRgb(h, s, l) {
-			var r, g, b;
-			if (s === 0) {
-				r = g = b = l;
+		function rgbToHsl(r, g, b) {
+			r /= 255, g /= 255, b /= 255;
+			var max = Math.max(r, g, b), min = Math.min(r, g, b);
+			var h, s, l = (max + min) / 2;
+			if(max === min) {
+				h = s = 0; // achromatic
 			} else {
-				var hue2rgb = function hue2rgb(p, q, t) {
-					if(t < 0) t += 1;
-					if(t > 1) t -= 1;
-					if(t < 1/6) return p + (q - p) * 6 * t;
-					if(t < 1/2) return q;
-					if(t < 2/3) return p + (q - p) * (2/3 - t) * 6;
-					return p;
-				};
-				var q = l < 0.5 ? l * (1 + s) : l + s - l * s;
-				var p = 2 * l - q;
-				r = hue2rgb(p, q, h + 1/3);
-				g = hue2rgb(p, q, h);
-				b = hue2rgb(p, q, h - 1/3);
+				var d = max - min;
+				s = l > 0.5 ? d / (2 - max - min) : d / (max + min);
+				switch(max) {
+				case r: h = (g - b) / d + (g < b ? 6 : 0); break;
+				case g: h = (b - r) / d + 2; break;
+				case b: h = (r - g) / d + 4; break;
+				}
+				h /= 6;
 			}
-			return [Math.round(r * 255), Math.round(g * 255), Math.round(b * 255)];
+			return [h, s, l];
 		}
-
-		var hash = input.split('-').join('');
-		var result = 0;
+		var hash = input.toLowerCase().replace(/[^0-9a-f]+/g, '');
+		var result = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+		var rgb = [0, 0, 0];
 		var sat = 80;
 		var lum = 68;
+		var modulo = 16;
 		for(var i in hash) {
-			result += parseInt(hash.charAt(i), 16)/16;
+			result[i%modulo] = result[i%modulo] + parseInt(hash.charAt(i), 16).toString();
 		}
-		result = result * 360;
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
Please add Screenshots before and after. THX a lot

skjnldsv:
Well, I changed my mind, I was exploring a new way to improve the color generator, but this doesn't seems good enough.
The current algorithm displays poor data repartition: (range is 0-360)
image 1

I tried working a second one, but can't figure out some issues in the data:
image 2
image 3

skjnldsv:
Okay, I managed to get something much more usable:
image 5

@DeepDiver1975 The colors are the same, I just changed the way the script transform the hash into a 0-360 int, so it has better random "distribution"

Henni:
I like these graphs!
@skjnldsv could you please document your thoughts/algorithm in the code

skjnldsv:
Hum, I can try @Henni! English not my main language but whatever! :)

The uuid system uses md5 which already is correctly distributed see [here, someone already did the job](http://michiel.buddingh.eu/distribution-of-hash-values)
The main goal was figuring out an algorithm which isn't heavy and gives us data that are evenly distributed between 0 and 360 (deg for the colours wheel).

At first i tried separating every char of the hash and converting form base16 to base10. Adding it up alltogether and dividing by the max we could get (32*16, md5 has 32char).
But because it's a sum, we have a function who converge in the medium values (see first graph).
So I came up with the idea to divise the hash in part, and remove the sum. (the modulo part on the script)

I decided to set up a formula that isn't mathematically logic, because I don't sum the base16 to 10. I convert the result to string and concatenate them, so 15+16 = 1516 and not 31.
The repartition overall should be even because all characters within the hash are 0-9a-f so we can have a statistically correct repartition.
BUT, I still needed to convert my hash division sum into a color. So to smooth the data even more, I decided to add the previous hash division results up within a modulo 3 that i could distribute to 3 variables: red, green and blue.

Then I just had to use the formula to convert rgb to hsl and retrieve the hue variable from the hsl.
Here is a beautiful paint to resume my terrible explanation! (The first letters are not in order, the characters position in the following graph are 1, 17, 2, 18, 3, 19, 4, 20.....)
fer

skjnldsv:
I'm sure the algorithm can be improved, but hey, do we really need a perfect colour generator? X)
I love working on those kinds of functions, but I'm not convinced this is vital ^^

skjnldsv:
We should probably cancel this PR and wait for this to be merged: https://github.com/owncloud/core/pull/23638

After we should update our templates to use the `element.imageplaceholder(hash,name)` function to generate our avatars

DeepDiver1975:
unit tests would be great to improve code coverage

skjnldsv:
I don't get what code coverage means. I found some website explaining how it works, or the purpose of this, but nothing actually helped me understand :)

skjnldsv:
I'm closing this because we need to use the core function.
refs https://github.com/owncloud/core/pull/23638

PR #314, opened 2016-03-29
Auto-resizing input for org and title layout + header improvements

skjnldsv:
- Set the fullName h2 and org+title div at 70% for better rendering
- Added a script to auto-resize the org & title inputs
- Added a max and min width on those inputs (20% & 50%) because they still are on the same line

This should allow for a clean display and something pretty good to see.
Mobile is good too, but of course the ellipsis are visible because of the low width-res.

See #293 and #286

### DIFF
--- a/l10n/es.js
+++ b/l10n/es.js
@@ -2,6 +2,7 @@ OC.L10N.register(
     "contacts",
     {
     "Contacts" : "Contactos",
+    "Address book name" : "Nombre de libreta de direcciones",
     "No contacts in here" : "No hay contactos aquí",
     "Name" : "Nombre",
     "Organization" : "Organización",
@@ -9,6 +10,7 @@ OC.L10N.register(
     "Add field ..." : "Añadiendo campo ...",
     "Import" : "Importar",
     "Add contact" : "Añadir contacto",
+    "No search result for {query}" : "Sin resultados para {query}",
     "All contacts" : "Todos los contactos",
     "Not grouped" : "No agrupado",
     "Post Office Box" : "Buzón de correo",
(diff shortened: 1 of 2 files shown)


jancborchardt:
Seems reasonably cool! :) I just wonder if the computation is too heavy cc @MorrisJobke @Henni 

Also, I would reduce the distance between the org and title input a bit (that is, the extra width the org field gets) cause currently they are a bit far apart.

skjnldsv:
This is due to the font width and the way inputs behave with the size attribute! :)
We can try to lower that tho!

jancborchardt:
Yeah, that’d be cool if you can do that :)

skjnldsv:
Updated with small size fixes and a typo.
@jancborchardt 
contacts owncloud

skjnldsv:
Rebased and ready to merge.

skjnldsv:
If we want to get the exact width of the text inside th einput, we might want to use a hack like this: http://plnkr.co/edit/DeoY73EcPEQMht66FbaB?p=preview

Get the text, put it on a div, get the div width, set the input to the same width and delete the temp div.
But it more heavier and dirty! :/
Just putting this here in case of future reference, I don't think we should bother doing this.

skjnldsv:
@tomneedham @DeepDiver1975 Should we merge? Or cancel and merge #293 instead?

jancborchardt:
@skjnldsv this is definitely better than #293 I think. :)

:+1:  just needs to be rebased again.

skjnldsv:
What have I done X)

skjnldsv:
Ahahahaha, i'm sooo done with git!
@jancborchardt HELP

Henni:
@skjnldsv did you do anything on github or did you just try to rebase?

skjnldsv:
I tried to rebase...
Please don't laugh... :'(

Henni:
can you recall which commands you executed?

skjnldsv:
`git rebase master`, I didn't had any conflict, so I pushed. i don't get it, I did it before on the mobile branch and I didn't get any trouble.

PR #321, opened 2016-03-30
fix app order

jancborchardt:
The idea is to have the sorting start with: Files, Activity, Gallery, Mail, Contacts, Calendar, and then the rest etc.

Please review @Henni @DeepDiver1975


DeepDiver1975:
:+1:

PR #330, opened 2016-04-01
Hide delete function for last email, phone and address field

### LINKED ISSUE #167: Hide delete function for last email, phone and address field
When there is only one field for email, phone, or address, the delete function should be hidden. Just as you can not remove the name, it makes more sense to permanently have them there for structure, even when not filled in.

skjnldsv:
Hide the delete button when there's only one field left for address, mails and phone, fix #167

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -152,6 +152,13 @@ detailsitem .icon-delete {
 	opacity: .2;
 }
 
+/* Prevent delete for last adr/mail/tel item  */
+.last-details > detailsitem.details-item-adr .icon-delete,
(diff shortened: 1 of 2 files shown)


jancborchardt:
Good stuff! Works nicely :+1:

skjnldsv:
This doesn't prevent the backend function tho, only the frontend by hiding the delete button.
Not that big of a deal. :)

Henni:
We might want to still show this icon. But instead of deleting the input it should just clear it.

jancborchardt:
Then we should only show it when the field is focused. Because otherwise it again clutters the view.

PR #331, opened 2016-04-01
Prevent change of contact-color when changing a contacts adressbook, fix #243

### LINKED ISSUE #243: Prevent change of contact-color when changing a contacts adressbook

skjnldsv:
Prevent change of contact-color when changing a contacts adressbook, fix #243

### DIFF
--- a/js/services/contact_service.js
+++ b/js/services/contact_service.js
@@ -76,10 +76,15 @@ angular.module('contactsApp')
 		}
 	};
 
-	this.create = function(newContact, addressBook) {
+	this.create = function(newContact, addressBook, uid) {
 		addressBook = addressBook || AddressBookService.getDefaultAddressBook();
 		newContact = newContact || new Contact(addressBook);
-		var newUid = uuid4.generate();
+		var newUid = '';
+		if(uuid4.validate(uid)) {
+			newUid = uid;
+		} else {
+			newUid = uuid4.generate();
+		}
 		newContact.uid(newUid);
 		newContact.setUrl(addressBook, newUid);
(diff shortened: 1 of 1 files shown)


jancborchardt:
@skjnldsv as pull request title and commit message, please always describe what was done. :) You can add the `fix issuenumber` at the end so it will be referenced, but it should always be human-readable / understandable.

jancborchardt:
@aspdye please review this as you opened the original issue :)

js/services/contact_service.js
```diff
@@ -76,10 +76,13 @@ angular.module('contactsApp')
 		}
 	};
 
-	this.create = function(newContact, addressBook) {
+	this.create = function(newContact, addressBook, uid) {
 		addressBook = addressBook || AddressBookService.getDefaultAddressBook();
 		newContact = newContact || new Contact(addressBook);
 		var newUid = uuid4.generate();
```
Henni: Please put this into an else clause. We don't have to do this if we have a valid uid.

skjnldsv:
Indeed, we could save a few resources!
I did it at first, but ended up defining it before because the tests failed and didn't allowed me to define twice Newuuid with var :)

skjnldsv:
Rebased!
There's a little flash when changing a contact addressbook, the delay between the contact deletion and the new contact creation.
The way we do this isn't maybe the right way, or could we stop the digest of angular between the delete and the new contact? I don't know. :/

DeepDiver1975:
:+1:

skjnldsv:
Thanks @DeepDiver1975

PR #339, opened 2016-04-05
Auto-resizing input for org and title layout + header improvements

skjnldsv:
REPLACE #314
- Set the fullName h2 and org+title div at 70% for better rendering
- Added a script to auto-resize the org & title inputs
- Added a max and min width on those inputs (20% & 50%) because they still are on the same line

This should allow for a clean display and something pretty good to see.
Mobile is good too, but of course the ellipsis are visible because of the low width-res.

Closes #293 and refs #286

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -12,27 +12,31 @@
 	top: 15px;
 	left: 100px;
 }
+.contactdetails__header #details-org-container {
+	position: absolute;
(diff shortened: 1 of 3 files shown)


skjnldsv:
Duplicate of #314. The branch got corrupted because of me, creating a new fresh pr ready to merge.

PR #341, opened 2016-04-06
Added contacts count in groups & fix group list not updating

skjnldsv:
- Added count with loading animation
- Fixed group list not updating when adding a new group

screencast

### DIFF
--- a/js/filters/counterFormatter_filter.js
+++ b/js/filters/counterFormatter_filter.js
@@ -1,0 +1,10 @@
+angular.module('contactsApp')
+.filter('counterFormatter', function () {
+	'use strict';
+	return function (count) {
+		if (count > 999) {
+			return '999+';
+		}
+		return count;
+	};
+});
--- a/js/filters/orderGroupList_filter.js
+++ b/js/filters/orderGroupList_filter.js
@@ -1,0 +1,27 @@
+angular.module('contactsApp')
+.filter('orderGroupList', function () {
+	'use strict';
+	return function (groups) {
+
+		// Init the order
+		var groupOrder = [t('contacts', 'All contacts'), t('contacts', 'Not grouped')];
+		var finalGroups = new Array();
+
+		// Converting groups object to Array and sort it
+		var groupArray = _.pairs(groups).sort();
+
+		// Replacing data or adding new one at the end (preserve the initial orger)
+		groupArray.forEach(function(group) {
+			var position = groupOrder.indexOf(group[0]);
+			var data = [group[0], group[1]];
+			// If in groupOrder, then insert at same pos
+			if(position !== -1) {
+				finalGroups.splice(position, 0, data);
+			} else {
+				// Add it at the end (we already sorted)
+				finalGroups.push(data);
+			}
+		});
+		return _.object(finalGroups);
+	};
+});
--- a/js/tests/filters/counterFormatter_filter.js
+++ b/js/tests/filters/counterFormatter_filter.js
@@ -1,0 +1,19 @@
+describe('counterFormatter filter', function() {
+
+	var $filter;
+
+	beforeEach(module('contactsApp'));
+
+	beforeEach(inject(function(_$filter_){
+		$filter = _$filter_;
+	}));
+
+	it('should return the same number or 999+ if greater than 999', function() {
+		var counterFormatter = $filter('counterFormatter');
+		expect(counterFormatter('15')).to.equal('15');
+		expect(counterFormatter('0')).to.equal('0');
+		expect(counterFormatter('-5')).to.equal('-5');
+		expect(counterFormatter('999')).to.equal('999');
+		expect(counterFormatter('1000')).to.equal('999+');
+	});
+});
--- a/js/tests/filters/counterFormatter_filter.js
+++ b/js/tests/filters/counterFormatter_filter.js
@@ -10,10 +10,11 @@ describe('counterFormatter filter', function() {
 
 	it('should return the same number or 999+ if greater than 999', function() {
 		var counterFormatter = $filter('counterFormatter');
-		expect(counterFormatter('15')).to.equal('15');
-		expect(counterFormatter('0')).to.equal('0');
-		expect(counterFormatter('-5')).to.equal('-5');
-		expect(counterFormatter('999')).to.equal('999');
-		expect(counterFormatter('1000')).to.equal('999+');
+		expect(counterFormatter(Number.NaN)).to.be.Nan;
+		expect(counterFormatter(15)).to.equal(15);
+		expect(counterFormatter(0)).to.equal(0);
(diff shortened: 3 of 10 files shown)


jancborchardt:
Nice!

Only one thing: there doesn't need to be a loading icon for every single counter, that's too distracting. Only a centered spinner in the app navigation as a whole.

skjnldsv:
Got it :)

Edit: done @jancborchardt

DeepDiver1975:
As already pointed out in one of these tickets - this feature will be thrown away once we move away from the current mechanism where we load all contacts initially (where displaying the count is easy) to a lazy loading mechanism - in lazy loading we have no possibility to get the total number of contacts in one group.

As a result we will have an immediate feature regression once we change the loading mechanism.

skjnldsv:
Too bad, I like this feature. :/
But it should be possible with addressbooks nonetheless right?

DeepDiver1975:
> Too bad, I like this feature. :/

I'm already having a look if there is a special query we could use for this case ...

skjnldsv:
Because to be fair, I rather have the addressbooks on the left than the groups, much better! :)

js/filters/counterFormatter_filter.js
```diff
@@ -0,0 +1,10 @@
+angular.module('contactsApp')
+.filter('counterFormatter', function () {
```
DeepDiver1975, line 2: this filter can easily be unit tested @skjnldsv - please add unit tests - thx

js/filters/orderGroupList_filter.js
```diff
@@ -0,0 +1,27 @@
+angular.module('contactsApp')
+.filter('orderGroupList', function () {
```
DeepDiver1975, line 2: this filter can easily be unit tested @skjnldsv - please add unit tests - thx

DeepDiver1975:
An alternative approach to pure lazy loading could be a mechanism where we build a full carddav client which uses webdav sync - a still uncertain question is where to put all the data in the browser thinking about really really big addressbooks.

skjnldsv:
I don't think loading all the data is a good idea. Right now, that's what we are doing and in my oc instance, i have 250+ contacts. The total weight is approximatively 20mo. I can't use the contact app on my mobile since the last update because of this insane amount of data I'm required to load on start. on wifi that's fine, but on 3g, that's terrible!

We should definitely go lazy and load only the data we need.

js/filters/counterFormatter_filter.js
skjnldsv, line 2: Hum, how?
Edit: ok I see, I will try to add a good test unit.

Henni, line 2: @skjnldsv Take a look at https://github.com/owncloud/contacts/blob/master/js/tests/filters/firstCharacter_filter.js

skjnldsv, line 2: I did, thanks :)

skjnldsv:
@DeepDiver1975 Here you go :)

js/tests/filters/counterFormatter_filter.js
```diff
@@ -0,0 +1,16 @@
...
+	beforeEach(module('contactsApp'));
+
+	beforeEach(inject(function(_$filter_){
+		$filter = _$filter_;
+	}));
+
+	it('should return the same number or 999+ if greater than 999', function() {
+		var counterFormatter = $filter('counterFormatter');
+		expect(counterFormatter('15')).to.equal('15');
+		expect(counterFormatter('1000')).to.equal('999+');
```
Henni: other interesting test cases:
`999`, `0`, `-5`

skjnldsv: Is this really necessary ?

js/tests/filters/orderGroupList_filter.js
```diff
@@ -0,0 +1,34 @@
...
+		['AGroup 8','0'],
+		['FGroup c','0'],
+		['Group 2','0'],
+		['Group 5','0'],
+		['Test 5','0']
+	);
+
+	it('should return the endGroupList order', function() {
+		var orderGroupList = $filter('orderGroupList');
+		expect(_.pairs(orderGroupList(startGroupList)).toString()).to.equal(endGroupList.toString());
```
Henni: Try `expect(orderGroupList(startGroupList)).to.deep.equal(endGroupList);`

skjnldsv: Thanks :)

jancborchardt:
Just for the record: Address books on the left instead of groups will definitely not happen. The most common usecase is just having one address book and not even knowing what an address book is. Groups however are commonly used in other apps and mobile operating systems (which we sync with) as well.

skjnldsv:
:'(

All in favor to merge?

jancborchardt:
Depends on what @DeepDiver1975 and @Henni say regarding the loading mechanism.

skjnldsv:
And depends if we're planning on changing the mechanism for 1.2 or after.

DeepDiver1975:
Introducing a feature we will remove sounds strange to me ....

js/tests/filters/counterFormatter_filter.js
```diff
@@ -0,0 +1,19 @@
...
+
+	beforeEach(module('contactsApp'));
+
+	beforeEach(inject(function(_$filter_){
+		$filter = _$filter_;
+	}));
+
+	it('should return the same number or 999+ if greater than 999', function() {
+		var counterFormatter = $filter('counterFormatter');
+		expect(counterFormatter('15')).to.equal('15');
```
DeepDiver1975: Testing NaN?

skjnldsv: Arrrgh! :D

skjnldsv:
Well, depends on what we decide, if we do a backend to handle dav, then we should be able to get the groups count even with lazy loading. If we don't, then we should drop this pr.

jancborchardt:
@Henni @DeepDiver1975 make the call.

From my side, it would be cool to have this. But only _after_ we accomodated the backend to deal with this. It does indeed not make sense to implement it to then drop it in a future version. In fact that would result in a very bad experience.

Henni:
We won't implement a php backend inside of the contacts app.
Maybe there will be an option to request the count directly from the backend in core, but I think we should close this for now.

skjnldsv:
Inside contacts no. But adding the functionality to dav why not?
Anyway, not having the ability to have your contact count is a real issue.

jancborchardt:
@skjnldsv yeah, maybe best open an issue about it in core so we can make a plan there. Or is there already one @DeepDiver1975?

DeepDiver1975:
No issue up to now - I'm uncertain about this since this is off the spec and I want to reduce custom implementations to an absolute minimum ....
... maybe we extend the groups query to return the count as well ... not too sure about this ....

PR #343, opened 2016-04-06
Fix header name input width

skjnldsv:
When creating new contact, from
screen1
to
screen2


jancborchardt:
Looks good :+1: 

Please also review @owncloud/designers

skjnldsv:
This is a simple fix, don't really think we need that many reviews :p

skjnldsv:
:+1:

Issue #337, opened 2016-04-05
Background of Header must no be white


ionum:
### Expected behaviour

Background of VCARD-Header must not be white
### Actual behaviour

If UID (which seems to be used for calculating the background color) contains an mail-like value (eg 20140426T014914.e765f629e7@domainname.net) background-color is white, which is bad for readability...
### Server configuration

ownCloud version:9.0.0
Contacts version: 1.1.0.0

skjnldsv:
Duplicate  #300
See #306

We're waiting for https://github.com/owncloud/core/pull/23768 to be merged so we can use the core function and fix this issue.

PR #345, opened 2016-04-07
Updated color filter to use the core function

### LINKED ISSUE #300: Invisible contacts letter avatar

skjnldsv:
All in the title! :)
After the merging of https://github.com/owncloud/core/pull/23768, we should now use the core function:
`input.toHsl()`

Fixes #300 #337 and ref #306.

cc @IwishIcanFLighT @ionum @bohni @macmeck please test with the core function.

### DIFF
--- a/js/filters/contactColor_filter.js
+++ b/js/filters/contactColor_filter.js
@@ -1,41 +1,17 @@
 angular.module('contactsApp')
 .filter('contactColor', function() {
 	return function(input) {
-		function hslToRgb(h, s, l) {
-			var r, g, b;
-			if (s === 0) {
-				r = g = b = l;
-			} else {
-				var hue2rgb = function hue2rgb(p, q, t) {
-					if(t < 0) t += 1;
-					if(t > 1) t -= 1;
-					if(t < 1/6) return p + (q - p) * 6 * t;
-					if(t < 1/2) return q;
-					if(t < 2/3) return p + (q - p) * (2/3 - t) * 6;
-					return p;
-				};
-				var q = l < 0.5 ? l * (1 + s) : l + s - l * s;
-				var p = 2 * l - q;
-				r = hue2rgb(p, q, h + 1/3);
-				g = hue2rgb(p, q, h);
-				b = hue2rgb(p, q, h - 1/3);
-			}
-			return [Math.round(r * 255), Math.round(g * 255), Math.round(b * 255)];
+		// Check if core has the new color generator
+		if(typeof input.toHsl === 'function') {
+			var hsl = input.toHsl();
+			return 'hsl('+hsl[0]+', '+hsl[1]+'%, '+hsl[2]+'%)';
+		} else {
+			// If not, we use the old one
+			/* global md5 */
+			var hash = md5(input).substring(0, 4),
+				maxRange = parseInt('ffff', 16),
+				hue = parseInt(hash, 16) / maxRange * 256;
+			return 'hsl(' + hue + ', 90%, 65%)';
 		}
-
-		var hash = input.split('-').join('');
-		var result = 0;
-		var sat = 80;
-		var lum = 68;
-		for(var i in hash) {
-			result += parseInt(hash.charAt(i), 16)/16;
-		}
-		result = result * 360;
-		var rgb = hslToRgb(result, sat, lum);
-		var bright = Math.sqrt( 0.299 * Math.pow(rgb[0], 2) + 0.587 * Math.pow(rgb[1], 2) + 0.114 * Math.pow(rgb[2], 2) );
-		if (bright >= 200) {
-			sat = 60;
-		}
-		return 'hsl('+result+', '+sat+'%, '+lum+'%)';
 	};
 });


DeepDiver1975:
Will this work with 9.0.0 as well?

skjnldsv:
It will only work with the updated placeholder generator from core, soooo it depends if they backlog it in 9.0 or not.

bohni:
Hi, works for me. Thanks!
Edit:
I just replaced this one line (without updating core)
`var hash = input.toHsl();`

I'm running 9.0.0 (stable)

Edit2:
Hm, should have read the whole thing... input.toHsl() gives me different colors for my contacts. No contact gets a white background anymore.
But I don't know if this gives me trouble on other circumstances...

bohni:
OK, little update
I now 
1) replaced `core/js/placeholder.js`completly with the raw file from https://raw.githubusercontent.com/owncloud/core/28571e6361eeb60ff241f8d8cc6a9a7e7fcdd318/core/js/placeholder.js
2) changed `contacts/js/public/script.js` lines 1584ff to

```
angular.module('contactsApp')
.filter('contactColor', function() {
        return function(input) {
                var hsl = input.toHsl();
                return 'hsl('+hsl[0]+', '+hsl[1]+'%, '+hsl[2]+'%)';
        };
});
```

I hope these are the right changes to do.
They work for me on 9.0.0 (stable)

Thanks!

skjnldsv:
Perfect!

DeepDiver1975:
as I was expecting - this is no working on a plain oc9

:-1:

bohni:
@DeepDiver1975 It is working, as I wrote in my first comment. But you can't be sure, that there are no side effects.

DeepDiver1975:
It is not working if you use a plain 9.0.0 as relased 4 weeks back because toHsl() simply does not exist.

skjnldsv:
Deep is right, you need the last placeholder.js update.
Should we plan this for 2.0 then ?

DeepDiver1975:
> Should we plan this for 2.0 then ?

does not really help - I want contacts to operate on a plain 9.0.0.

We need to add a fallback in case toHsl is not provided by core

skjnldsv:
And you're perfectly right! 
The last 9.0 update from core doesn't even include my first color generator update. I don't exactly get how the roadmap is defined. We should get back to the same old function placeholder.js is using (since this is what is shipper with 9.0.x) and when core will be upgraded with my last color generator, we should add those two lines in this and keep the old one just in case.

This should allow retro-compatibility and should not make this package heavier by adding another color generator. What do you think?

skjnldsv:
This should do the trick, and we would have the same colors than core (like @jancborchardt wanted in the first place)

```
if(typeof input.toHsl === 'function') {
    var hsl = input.toHsl();
    return 'hsl('+hsl[0]+', '+hsl[1]+'%, '+hsl[2]+'%)';
} else {
    var hash = input.substring(0, 4),
        maxRange = parseInt('ffff', 16),
        hue = parseInt(hash, 16) / maxRange * 256;
    return 'hsl(' + hue + ', 90%, 65%)';
};
```

bohni:
@skjnldsv 
if you use `var hash = input.substring(0, 4),` instead of your solution in #300 (`var hash = input.toLowerCase().replace(/[^0-9a-f]+/g, '');`) all contacts will get the same color, if the UID starts with the same timestamp as the one I showed in #300 (`20131217T100925.ac9f58a2cd@example.com`)

skjnldsv:
Yeah, I forgot a md5 :)

bohni:
@DeepDiver1975 @skjnldsv 
I just downloaded and extracted 
- owncloud-9.0.0.tar.bz2 and copied the file `core/js/placeholder.js` to my running 9.0.0 instance
- contacts.tar.gz (1.1.0.0  from https://apps.owncloud.com/content/show.php/Contacts?content=168708) and replaced my changed js/public/script.js  on my running 9.0.0 instance

This leaded to my contacts background getting white again.

Then I changed line 1609 from
`var hash = input.split('-').join('');`
to
`var hash = input.toHsl();`
and my contacts backgorund became colored again

The diff between the extracted owncloud.tar and my instance shows:

```
$ diff -r oc owncloud-9
Only in oc/apps: calendar
Only in oc/apps: contacts
Only in oc/config: config.php
Only in oc/config: config.php.orig
```

So the change is working on plain 9.0.0

Edit: Maybe that the colors are not as randomized as with the complete change

DeepDiver1975:
> So the change is working on plain 9.0.0

only because you patched placeholder.js - which is wrong

DeepDiver1975:
> This should do the trick, and we would have the same colors than core (like @jancborchardt wanted in the first place)

@skjnldsv exactly what we should do! THX

bohni:
@DeepDiver1975
I wrote, that I've undone the changes and the diff shows no difference. BUT

@DeepDiver1975  @skjnldsv 

Sorry guys. My fault. Cache or something fooled me. I used Shift-F5 to reload and colors are shown. I again shift-f5 reloaded the page and colors are gone.

skjnldsv:
No pb :)

skjnldsv:
Rebased and ready to merge.
@DeepDiver1975 compatibility is ok now.

Test procedure:
1. replace /core/js/placeholder.js with https://github.com/owncloud/core/blob/stable9/core/js/placeholder.js and check if you get the old ugly colours
2. now replace with https://github.com/owncloud/core/blob/master/core/js/placeholder.js and check if you get the clean "pastel" colours

skjnldsv:
@jancborchardt could you take a look?

cc @bohni @IwishIcanFLighT @ionum @macmeck @criegerde please review following this comment https://github.com/owncloud/contacts/pull/345#issuecomment-207238351

ghost:
Results as shown attached:

core: 
core
master:
master

jancborchardt:
@criegerde @skjnldsv so it results in different colors? Shouldn’t they be the same? The algorithm shout definitely spit out the same color for the same value put in.

ghost:
yes, it results in different colors.

skjnldsv:
@criegerde no there is a problem, the first one should have terrible colors like this: 
screen
And with the new generator:
screen2

Did you applied my commit?

@jancborchardt No, this is normal, the last core update (9.0) does not include the new generator, so we're including the old color gen from core inside contacts. But we also added an alternative to be ready when core will have the last generator. Like so, we always have a matching color between core and contacts, regardless of the core version and whatever color generator the user has installed. :)

ghost:
@skjnldsv yes, i did:
1. replace /core/js/placeholder.js with https://github.com/owncloud/core/blob/stable9/core/js/placeholder.js  **<= called core**
2. now replace with https://github.com/owncloud/core/blob/master/core/js/placeholder.js **<= called master**

skjnldsv:
And what about the code in contacts?

ghost:
what do you mean exactely?

skjnldsv:
in contacts/js/public/script.js, search the code `filter('contactColor'`

And replace the whole section by this: https://github.com/owncloud/contacts/blob/use-core-colors-generator/js/filters/contactColor_filter.js
(the old function should be visible here: https://github.com/owncloud/contacts/blob/8fc233da68b2e8884d0097bddfb2ca9d87177679/js/filters/contactColor_filter.js)

ghost:
apologize, i forgot this step.
but unfortunately the colors are still different 


positiv: the empty/white/blank letter-Icon changed to a colored one - the name of this contact is also readable

skjnldsv:
Yeah, the colors are supposed to be different, that's normal :)

Thank you for your feedback! :+1:

ghost:
ok, so it works fine for me!
i checked another ownc9.0.1-instance - same positiv results with clean pastel colors.
thank you!

bohni:
Hello,

here my results:
bugfix-result

left side: old core
right side: new core

top: old contacts
center: contacts with fix from #300 
bottom: new contacts

Bohni

skjnldsv:
So to be clear the bottom line is with this code?
https://github.com/owncloud/contacts/blob/use-core-colors-generator/js/filters/contactColor_filter.js

(This is confusing enough with all those colors, I want to be sure :p )

bohni:
> So to be clear the bottom line is with this code?

Yes, with "new contacts" I meant your new code your link points to

skjnldsv:
Perfect!

skjnldsv:
@DeepDiver1975 could you do a last check? :)

DeepDiver1975:
:+1:

PR #346, opened 2016-04-07
Hide the Add Contact button if we've got an invalid contact

### LINKED ISSUE #160: When a new contact has no details at all filled in yet, it should not be possible to add another one
Right now you can just add empty contacts endlessly, which makes little sense.

When a new contact has no specific info filled in yet, clicking »Add new contact« should just focus the name field of the already added one.

skjnldsv:
This prevent from adding multiple empty contacts as long as you don't have filled the full name field (aka FN).
- Hides the new contact button when we have an invalid contact
- Resized the button height to match a contactList item height.
- A new contact now take the place of the button.

This fixes #160
screencast

### DIFF
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -7,6 +7,7 @@ angular.module('contactsApp')
 	ctrl.contactList = [];
 	ctrl.searchTerm = '';
 	ctrl.show = true;
+	ctrl.invalid = false;
 
 	ctrl.t = {
 		addContact : t('contacts', 'Add contact'),
@@ -146,6 +147,15 @@ angular.module('contactsApp')
 				}
 				unbindWatch(); // unbind as we only want one update
 			});
+		}
+	});
+
+	// Watch if we have an invalid contact
+	$scope.$watch('ctrl.contactList[0].fullName()', function(fullName) {
+		if(fullName === '') {
+			ctrl.invalid = true;
+		} else {
+			ctrl.invalid = false;
 		}
 	});
 
--- a/js/components/contactList/contactList_controller.js
+++ b/js/components/contactList/contactList_controller.js
@@ -152,11 +152,7 @@ angular.module('contactsApp')
 
 	// Watch if we have an invalid contact
 	$scope.$watch('ctrl.contactList[0].fullName()', function(fullName) {
-		if(fullName === '') {
-			ctrl.invalid = true;
-		} else {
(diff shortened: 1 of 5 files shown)


js/components/contactList/contactList_controller.js
```diff
@@ -145,6 +146,15 @@ angular.module('contactsApp')
 		}
 	});
 
+	// Watch if we have an invalid contact
+	$scope.$watch('ctrl.contactList[0].fullName()', function(fullName) {
```
DeepDiver1975, line 154: Is it always the first contact in the list?
Looks strange to me

js/components/contactList/contactList_controller.js
```diff
@@ -145,6 +146,15 @@ angular.module('contactsApp')
...
+		if(fullName === '') {
+			ctrl.invalid = true;
```
DeepDiver1975: Can we a one liner ...

js/components/contactList/contactList_controller.js
skjnldsv, line 154: if the fn field is empty, and therefore the contact invalid, then yes, always the first one on the list.

js/components/contactList/contactList_controller.js
skjnldsv: Done

jancborchardt:
Only disable the button instead of hiding it. That will prevent the list from making this strange jump.

skjnldsv:
The button hiding is exactly to prevent the list from making the jump, it replaces the button with the same height and therefore does not move your whole list to the bottom. I though this was a good idea.

jancborchardt:
Actually good point, yeah. Let's do it like that.

I have some ideas on how to make the »Add contact« function look more nicely in the list than just a button too.

jancborchardt:
Very nice!

Two small enhancements:
- until a proper name is put in, show »New contact« in the first line of that newly created contact
- when clicking on another contact, and then back on that newly created contact (without any info put in yet), the full name field should automatically be focused

skjnldsv:
Good ideas!

skjnldsv:
New contact is now showed. For the focus I'm having trouble, it does not seems to work. I tried adding 

```
if(ctrl.contact.fullName() === '') {
    $('#details-fullName').focus();
}
```

On contactsDetails_controller.js in the `ContactService.getById(uid).then(function(contact) {` line.
Doesn't seems to work.

DeepDiver1975:
:+1:

PR #323, opened 2016-03-31
First contact load fix, direct access to contact, contact not in group, group load and order - fix #312

skjnldsv:
Fix multiple issues:
- When loading the page directly on a specific group, the first contact loaded was taken from the main contact list which often results by displaying a contact not in the requested group
- When reclicking the already selected group from the list after a fresh page load resulted by the contact details disappearing
- The contacts starting with a special character was being ignored when choosing the first contact to load
- When a uid is specified, the script will  now be looking for it in the specified group, if not found, the first contact of group will be displayed

PR #347, opened 2016-04-07
Fix error loading page when no contacts

skjnldsv:
Did a mistake with the new loading fix (ref #323)
When you don't have any contacts, loading the page generate an error and doesn't load further.


DeepDiver1975:
:+1:

PR #348, opened 2016-04-08
fix delete button alignment of address fields delete button

jancborchardt:
The icon somehow moved into the P.O. box field … should be right of the whole block, so that’s fixed now.

Please review @Henni  @skjnldsv @irgendwie


DeepDiver1975:
:+1:

Issue #248, opened 2016-03-20
Export Addressbook


Henni:
Add option to export an addressbook with all its contacts.

jancborchardt:
Yes! This is quite important regarding our principle of data portability.

DeepDiver1975:
The Backend is already providing this feature.
We need to trigger the download only in the web ui.

skjnldsv:
@DeepDiver1975 do you know how the backend is providing this? What is the url for the download?
The contacts are https://domain.ext/remote.php/dav/addressbooks/users/USER/ADDRESSBOOK/UID.vcf, but can't figure out for addressbooks :)

DeepDiver1975:
the correct url is  https://domain.ext/remote.php/dav/addressbooks/users/USER/ADDRESSBOOK?export - but we need to fix core first - https://github.com/owncloud/core/pull/23893

skjnldsv:
You're awesome! :)
:+1:

DeepDiver1975:
addressbook export will be part of 9.0.2 - https://github.com/owncloud/core/pull/23897

@skjnldsv we better have a version check to see if addressbook export is enabled or not

skjnldsv:
Okay! 
On it!

captainwasabi:
I need this too!!! I have applied owncloud/core#23897 by hand.  Waiting for client side.

Oops, never mind.  Clicked the wrong button and deleted the whole address book! :(  Might I suggest perhaps a confirmation dialog before two days work of hand entering addresses is destroyed?  Oops, too late it's gone. ;)

skjnldsv:
It's planned  too!

PR #342, opened 2016-04-06
Change, delete and replace contacts avatar - fix #144

skjnldsv:
WIP request for the upload management of contacts profile pictures.
@Henni

PR #349, opened 2016-04-10
Export contact and addressbook as vcard + header restructuration fix

### LINKED ISSUE #247: Export Contact
Add option to export a single contact as a vCard

skjnldsv:
Fixes #247 and #248
Waiting for https://github.com/owncloud/core/pull/23891 for the white download icons
- Add option to export an addressbook with all its contacts
- Add option to export a single contact as a vCard

screen

### DIFF
--- a/js/components/addressBook/addressBook_controller.js
+++ b/js/components/addressBook/addressBook_controller.js
@@ -3,6 +3,10 @@ angular.module('contactsApp')
 	var ctrl = this;
 
 	ctrl.showUrl = false;
+	/* globals oc_config */
+	/* eslint-disable camelcase */
+	ctrl.canExport = oc_config.versionstring.split('.') >= [9, 0, 2];
+	/* eslint-enable camelcase */
 
 	ctrl.toggleShowUrl = function() {
 		ctrl.showUrl = !ctrl.showUrl;
--- a/templates/addressBook.html
+++ b/templates/addressBook.html
@@ -1,5 +1,12 @@
 <span class="addressBookName">{{ctrl.addressBook.displayName}}</span>
 <span class="utils">
+	<span class="action" ng-show="ctrl.canExport">
+		<span href="{{ctrl.addressBook.url}}?export"
+			  id="addressBooklist-icon download"
+			  title="Download"
+			  class="icon-download">
+		</span>
+	</span>
     <span class="action">
         <span
             class="addressbooklist-icon icon-public"
--- a/templates/addressBook.html
+++ b/templates/addressBook.html
@@ -1,11 +1,11 @@
 <span class="addressBookName">{{ctrl.addressBook.displayName}}</span>
 <span class="utils">
 	<span class="action" ng-show="ctrl.canExport">
-		<span href="{{ctrl.addressBook.url}}?export"
+		<a href="{{ctrl.addressBook.url}}?export"
 			  id="addressBooklist-icon download"
 			  title="Download"
 			  class="icon-download">
-		</span>
+		</a>
 	</span>
     <span class="action">
         <span
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -7,41 +7,43 @@
 	align-items: center;
 }
 
+.contactdetails__header #details-contact-infos {
+	width: 80%;
+	margin: 2px 6px 0px;
+	flex-grow: 1;
+}
 .contactdetails__header h2 {
-	position: absolute;
-	top: 15px;
-	left: 100px;
-	width: 75%;
+	margin: 0;
 }
 .contactdetails__header #details-org-container {
-	position: absolute;
 	display: flex;
-	top: 50px;
-	left: 100px;
-	width: 70%;
+	flex-wrap: wrap;
 }
 .contactdetails__header .contactdetails__name,
 .contactdetails__header .contactdetails__org,
 .contactdetails__header .contactdetails__title {
-	font-size: inherit !important;
-	width: auto !important;
-	color: #fff !important;
-	background: transparent !important;
-	border: none !important;
+	font-size: inherit;
+	color: #fff !important; /* override focus,active&hover */
+	background: transparent;
+	border: none;
 	text-overflow: ellipsis;
 	overflow: hidden;
 	white-space: nowrap;
 	max-width: 100%;
-	min-width: 25%;
-	margin: 3px 0;
+	margin: 0;
+    padding: 4px 5px;
 }
 .contactdetails__header .contactdetails__org,
 .contactdetails__header .contactdetails__title {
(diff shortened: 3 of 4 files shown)


jancborchardt:
The icons are too close together and look a bit strange on top of each other. I would place them left/right next to each other, with the delete icon on the right.

skjnldsv:
Like this?
Screen

jancborchardt:
Yes, but with more padding. The clickable area of an icon needs to be 44*44px at least.

skjnldsv:
Okay, and what about the mobile view?
It break a lot of stuff.
I think we need to fix the header, using position absolute in css isn't really a good idea for media queries.
screen

jancborchardt:
On mobile it could be shown above each other, it’s the less important view. But on desktop it should look proper.

> I think we need to fix the header, using position absolute in css isn't really a good idea for media queries.

What do you mean by that? The name should always be shown fixed? I would say not because then you have very little space for typing in the fields, when the keyboard is also expanded.

skjnldsv:
I mean, in css, you need to keep a structure.
Using the absolute position in css should only be used for elements that break the cascading structure of your page. In this case, the h2, inputs, buttons, don't break this at all.
We should use either a flexbox or a float, but using absolute break many further possibilities to make a dynamic design, because the element become an independent item and isn't related or linked to elements around it.

This is a good example, the header become broken in mobile view. With a good fluid css design, the elements should at least rearrange to display themselves in the page. Because of the absolute we got div out of the view! Not a good idea!

Unfortunately, a lot of design part in owncloud use the position absolute option and I don't think this is very convenient. :)

DeepDiver1975:
We have the recurring issue that this is not working with 9.0.0 and 9.0.1 because the icon is not in core.
Please add the download icon to the app and use this one until we drop support for 9.0.0 and 9.0.1 (assuming we backport owncloud/core#23891 )

skjnldsv:
Okay, I wasn't sure how to proceed! Thanks Deep! :)

jancborchardt:
@skjnldsv actually the absolute positioning in combination with the correct sizing (width percentage), testing and additional responsive rules works very well everywhere. Floats on the other hand are very difficult to control. Most of the time our layout is simple anyway (and I intend to keep it that way), with 3 elements max next to each other, so absolute positioning is fine.
Hope that explains it? :)

Henni:
@skjnldsv I suggest that this PR should only introduce the export of a single contact.
Let's leave the export function for addressbooks to a different PR as it seems to be a bigger task.

jancborchardt:
Yeah, however the export of the whole address book has higher priority for sure. I would say it should be done first.

skjnldsv:
@Henni this is a single link for the addressbook too. With just a condition to check the dav lib version.
Shouldn't be that hard.

@jancborchardt no, absolute positioning is not fine at all. It works because it's correctly written, but it's terrible coding for css! :p 
It's my humble opinion. And for the float, it's hard to master, but works well (even if now flex is faaar better for positioning and alignement, and should put float on the bench.). :)

For the priority I worked for things that was possible, I couldn't do the addressbook export as long as we didn't have the possibility. Now we do, commit on his way.

Edit: sorry if I sound presumptuous, this isn't on purpose at all, just my opinion (bad or good, you decide) (and i also just came back from work and I'm exhausted ;) )

DeepDiver1975:
> With just a condition to check the dav lib version.
> Shouldn't be that hard.

If I remember correctly we have no mechanism on the js side to read the version number ....

DeepDiver1975:
wait ... oc_config can be used ...

bildschirmfoto von 2016-04-12 09-43-47

skjnldsv:
Thanks @DeepDiver1975 :+1:

skjnldsv:
@DeepDiver1975 ok, seems to work. I don't really like using the controller as it seems easier to pass the whole thing into a ng-shown but parseFloat isn't declared into $scope so I can't use it, it occurred to me it was simpler this way. What do you think? Is this over complicated?

@jancborchardt I will update the header later :)

screen

skjnldsv:
@jancborchardt I fixed the whole header. This is now more flexible and cross-device. 
This is what growing and shrinking looks like:
untitled

ps: because there is no profile picture, the missing space is ok. This should be fixed when the upload form will be done by Henni.

jancborchardt:
@skjnldsv I do like the restructuring except for the removal of the left whitespace. On the most narrow width it’s ok, but on the other widths it’s intentional to line up with the field contents.

skjnldsv:
The left space is because i'm missing the div that replace the img in #342 :)
I can add it if you want?

jancborchardt:
Yeah, it should be added please – it gives better visual balance. You can remove it on the most narrow view if you like, but we need to show the avatar in some way there anyway.

skjnldsv:
No no, I agree with you! :)

skjnldsv:
Rebased and ready to go!

js/components/addressBook/addressBook_controller.js
```diff
@@ -3,6 +3,9 @@ angular.module('contactsApp')
 	var ctrl = this;
 
 	ctrl.showUrl = false;
+	/* globals oc_config */
+	/* eslint camelcase: [2, {properties: "never"}] */
+	ctrl.canExport = parseFloat(oc_config.version.substring(0, 3)) >= 9.1;
```
DeepDiver1975: float is not really the right thing to use - generally speaking since 9.1 and 9.10 are equal in terms of float but not in terms of version. Anyway - there will never be a version higher then 9.2.

BUT: 9.1 is not the correct value to test against. you need to test against 9.0.2 because the fix has been backported

skjnldsv: Oh okay! Thanks! :)
What should I do then?

skjnldsv:
@DeepDiver1975 rebased and fixed.

templates/addressBook.html
```diff
@@ -1,5 +1,12 @@
 <span class="addressBookName">{{ctrl.addressBook.displayName}}</span>
 <span class="utils">
+	<span class="action" ng-show="ctrl.canExport">
+		<span href="{{ctrl.addressBook.url}}?export"
```
DeepDiver1975: span? Doesn't work for me ....

skjnldsv: Do you have the version > 9.0.2?
It works fine here, I don't understand. What's wrong?
screen-34-04

skjnldsv: ... No comment, I didn't realized! ^^
I must have left some testing.

Thanks for the commit.

skjnldsv:
Rebased and waiting for tests before merging.

jancborchardt:
Nice, very cool!

Issue #70, opened 2016-03-01
remove emtpy group


DeepDiver1975:
as soon as the last contacts is deleted from one group we should remove the group from the group list

DeepDiver1975:
already fixed with #353

PR #353, opened 2016-04-11
Fixed group list not updating when adding/deleting a group

### LINKED ISSUE #319: Import doesn’t directly show the groups
When importing, contacts directly show. The groups in the left navigation bar only show up after a manual refresh though.

skjnldsv:
All in the title! :)
Testing:
1. Install commit
2. Click/create a contact and add a new group
3. Delete the group
4. Import a contact with a group

Expected: the group list should update itself.
Fix #319, #70


DeepDiver1975:
tested - works :+1:

PR #366, opened 2016-04-15
In case a homepage entry does not hold the protocol we append it to f…

DeepDiver1975:
…ix the link and the input field value


skjnldsv:
Hi deep! :)
Found some bugs I think:
1. When I set an invalid string, the input isn't actualised immediately
2. And when I switch to a card and go back to the one I edited, the field has just http before:
cap

Issue #365, opened 2016-04-15
Missing field types and values and different labels


TecJon:
### Steps to reproduce
1. Create a contact in iOS using different fields
2. Open contact in owncloud web interface.
### Expected behaviour

See the contact with the same labels and values.
### Actual behaviour

Some fields are missing or have different labels (field types):
- Social media fields are all missing
- Date "anniversary" is missing
- "mobile" type is called "Cell Voice" - but I believed this was mentioned before

![untitled](https://cloud.githubusercontent.com/assets/11375752/14561823/57d35f80-0319-11e6-9c9c-43086d781d1d.png)

![untitled2](https://cloud.githubusercontent.com/assets/11375752/14561864/a0e32caa-0319-11e6-8ec2-caa1cf47922d.png)
### Server configuration

**Operating system**:  Ubuntu 14.04

**Web server:** nginx

**Database:** MySQL

**PHP version:** PHP 5.5.9-1ubuntu4.14 (fpm-fcgi) (built: Oct 28 2015 01:38:24)

**ownCloud version:** 9.0.1.3

**Contacts version:** 1.1.0.0

**Updated from an older ownCloud or fresh install:** updated

**Signing status (ownCloud 9.0 and above):**
Long list due to failed integrity test.

```
Technical information
=====================
The following list covers which files have failed the integrity check. Please read
the previous linked documentation to learn more about the errors and how to fix
them.

Results
=======
...

```

**List of activated apps:**

```
Enabled:
  - activity: 2.2.1
  - announcementcenter: 1.1.1
  - calendar: 1.1
  - comments: 0.2
  - contacts: 1.1.0.0
  - dav: 0.1.6
  - documents: true
  - federatedfilesharing: 0.1.0
  - federation: 0.0.4
  - files: 1.4.4
  - files_pdfviewer: 0.8
  - files_sharing: 0.9.1
  - files_texteditor: 2.1
  - files_trashbin: 0.8.0
  - files_versions: 1.2.0
  - files_videoplayer: 0.9.8
  - firstrunwizard: 1.1
  - gallery: 14.5.0
  - galleryplus: true
  - notifications: 0.2.3
  - provisioning_api: 0.4.1
  - systemtags: 0.2
  - templateeditor: 0.1
  - updatenotification: 0.1.0
Disabled:
  - encryption
  - external
  - files_external
  - user_external
  - user_ldap
```

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc8ab27a8968",
[...]

PR #367, opened 2016-04-15
Adding social profile fields

DeepDiver1975:
Display social profile fields as described in https://github.com/owncloud/contacts/issues/365


skjnldsv:
Nice, 👍 
This also a planned feature for the vcard format: https://tools.ietf.org/id/draft-george-vcarddav-vcard-extension-02.html#rfc.section.2.2

PR #368, opened 2016-04-15
Don't allow to delete the last addressbook

### LINKED ISSUE #139: Last address book, should it be deletable?
Seems like the last address book should not be possible to delete. Contacts without an address book does not make sense, and this could happen mostly by accident.

You should only be able to delete an address book when there are at least two. When there’s a single address book, the delete icon should not be there.

DeepDiver1975:
@jancborchardt 

fixes #139

### DIFF
--- a/templates/addressBook.html
+++ b/templates/addressBook.html
@@ -17,6 +17,7 @@
     <span class="action">
         <span
             class="addressbooklist-icon icon-delete"
+			ng-class="{hidden: ctrl.list.length === 1}"
             title="Delete Addressbook"
             ng-click="ctrl.deleteAddressBook()">
         </span>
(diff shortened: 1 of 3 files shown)


templates/addressBook.html
```diff
@@ -17,6 +17,7 @@
     <span class="action">
         <span
             class="addressbooklist-icon icon-delete"
+			ng-class="{hidden: ctrl.list.length === 1}"
```
skjnldsv, line 20: Why not use the ng-show?

skjnldsv:
👍, works great!

jancborchardt:
Yep, works nicely! :+1:

PR #375, opened 2016-04-20
Fixed loader height

skjnldsv:
Simple & very small fix:
From 
capture d ecran_2016-04-20_07-50-33
To
capture d ecran_2016-04-20_07-50-43


DeepDiver1975:
:+1:

PR #379, opened 2016-04-20
In case no FN is defined we use N - temporal hack for now until we pr…

### LINKED ISSUE #322: Addressbook entry's not shown, but synchronized
I have a lot vcard entry's, which contain only the "N" line, the "FN" line is missing. I don't know, where they are coming from, maybe from an old "kaddressbook" app or from older owncloud version, or from android before, no way to find out anymore. If I actually add an address, the "FN" is there.
Anyway on owncloud prior version 9, this entry's was shown. On android this vcard entry's are shown, on "kaddressbook" this entry's are shown, on owncloud 9 this entry are not show up anymore.
### Steps to reproduce
1. create a vcard, open in text editor and delete "FN" line
2. import to owncloud
3. look in the db, the vcard is there, even updates over carddav are working, but not shown in contacts app
### Server configuration

**ownCloud version:**
9.0.0.19
**Contacts version:**
1.1.0.0
**Updated from an older ownCloud or fresh install:**
update

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```

No errors have been found.

```

```

DeepDiver1975:
…operly support N handling - fixes #322

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -33,9 +33,12 @@ angular.module('contactsApp')
 					var property = model.getProperty('fn');
 					if(property) {
 						return property.value;
-					} else {
-						return undefined;
 					}
+					property = model.getProperty('n');
+					if(property) {
+						return property.value.join();
+					}
+					return undefined;
 				}
 			},
 


skjnldsv:
:+1:

PR #383, opened 2016-04-21
Fix sorting

### LINKED ISSUE #382: Sorting issues after import
On a recent import, this happened:
capture du 2016-04-21 13-35-31 capture du 2016-04-21 13-27-27
Obviously there’s something wrong here … random contacts which start with »J« are sorted first in list, then someone with »Ar« who should be lower than someone with »An« and »Al«. The screenshots are from 2 different imports and persist after refresh.

DeepDiver1975:
fixes #382

@jancborchardt


jancborchardt:
Nice, seems to be fixed now! :+1:

PR #402, opened 2016-05-06
Import button 

### LINKED ISSUE #279: Import button overlaps settings button in german ui

skjnldsv:
Let's go back to the previous design. It's not pretty, but functional. If we ever need a secondary button, we could but it right next to it so it could be more beautiful.

fix #279

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -324,21 +324,15 @@ li.addressBook-share-item span.shareeIdentifier {
 }
 
 #app-settings-content #upload.button {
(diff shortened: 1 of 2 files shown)


skjnldsv:
capture d ecran_2016-05-06_18-00-07

DeepDiver1975:
:+1:

Issue #354, opened 2016-04-11
Split Contact Details into two colums


aspdye:
Otherwise, i think that it looks strange on larger screens :grin: 

cc @jancborchardt

jancborchardt:
Yes, definktely. We had that in the old app too. We can for sure put the big address blocks in the second column, then it's also consistent.

skjnldsv:
Agreed!
Like this?
![contacts owncloud](https://cloud.githubusercontent.com/assets/14975046/14585423/15c88176-0475-11e6-87c3-c3b5bf8cd7fe.png)

aspdye:
Oh no ... It was about the Details (phone, adress, ...) we want to split :grin:

skjnldsv:
Okay, I need to learn to read again...
Well, looked fun anyway :D

jancborchardt:
LOL :D Yeah, definitely not two rows of contacts.

skjnldsv:
![capture d ecran_2016-05-14_11-56-05](https://cloud.githubusercontent.com/assets/14975046/15267762/db8c5378-19ca-11e6-928e-77343b8c3bef.png)

Like this?
This needs a lot of work, and fix _(jan, stop using absolute positionning, it's everywhere !!!!)_ But it's working great :)

aspdye:
Great work :rocket: That was exactly what i have meant.

jancborchardt:
@skjnldsv nice! Just a few things:
- a bit more whitespace between the columns would be good
- I would put the one-liner stuff on the left (email, phone, etc) and the multiline post address on the right, so that it’s balanced

skjnldsv:
Yes, I haven't had the time to continue my work on it these days! :( 
I should be available soon. :)

manishbisht:
@jancborchardt @skjnldsv @aspdye @DeepDiver1975
Hi, I am working on this issue. I have split the contact details section into 2 columns if the screen width is greater then 1200px.

Now by default only 3 fields i.e Phone, Email and Address are shown and after splitting into 2 columns there is more empty space shown in the contact details section which looks bad.

Now how can I display the more fields like Website, Nickname, Notes etc using ng-repeat in the contact details section or the other way to display more fields will be to call the ctrl.addFiled(field) function.

Also I want to know that what parameters does the ctrl.addFiled function accepts.

jancborchardt:
What is very important regarding layout:
- use 2 columns, and not more!
- left column is phone, email address etc
- right column is addresses
[...]

PR #411, opened 2016-05-14
Add multiple row in contact details and fixes some css

skjnldsv:
ref #354

### DIFF
--- a/templates/contactDetails.html
+++ b/templates/contactDetails.html
@@ -31,11 +31,11 @@
 		</header>
 		<section>
 			<div ng-repeat="prop in ctrl.contact.props | toArray | orderDetailItems:'$key'">
-				<detailsItem ng-repeat="propData in prop" name="prop.$key" data="propData" model="ctrl" index="$index" ng-class="[ 'details-item-' + prop.$key ]"></detailsItem>
+				<detailsItem ng-repeat="propData in prop" name="prop.$key" data="propData" model="ctrl" index="$index" ng-class="[ 'details-item-' + prop.$key, {'hidden': 'org title version fn geo photo uid'.indexOf(prop.$key) !== -1} ]"></detailsItem>
 			</div>
-			<div class="select-addressbook" ng-if="ctrl.addressBooks.length > 1">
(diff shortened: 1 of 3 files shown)


aspdye:
:+1:

skjnldsv:
@aspdye still in progress, we need to make everythings more readable :)
Thanks!

skjnldsv:
Added the hidden class for the unwanted detailsitems elements (fn, title, org and version...), fixed delete button alignements.
And fixed the inputs heights in case of flex-grow :)

We still need to do something for clarity I think. Space them?
@Henni @jancborchardt @aspdye ?

templates/contactDetails.html
```diff
@@ -31,11 +31,11 @@
 		</header>
 		<section>
 			<div ng-repeat="prop in ctrl.contact.props | toArray | orderDetailItems:'$key'">
-				<detailsItem ng-repeat="propData in prop" name="prop.$key" data="propData" model="ctrl" index="$index" ng-class="[ 'details-item-' + prop.$key ]"></detailsItem>
+				<detailsItem ng-repeat="propData in prop" name="prop.$key" data="propData" model="ctrl" index="$index" ng-class="[ 'details-item-' + prop.$key, {'hidden': 'org title version fn geo photo uid'.indexOf(prop.$key) !== -1} ]"></detailsItem>
```
skjnldsv, line 34: I need help with this.
The best way is to access the detailsItem scope and check for ctrl.meta.template==='hidden'.
I can access it on the console through `.scope().$$childHead.ctrl.meta.template` But I can't in this template.

Any idea?
@DeepDiver1975 @Henni @owncloud/javascript

Henni, line 34: @skjnldsv What do you want to accomplish here?

skjnldsv, line 34: Add the hidden class for all detailsItem which ctrl.meta.template is set to hidden.

Henni, line 34: Why do you want to hide them here?
This is already done in the detailsItem template (hidden.html) isn't it?

skjnldsv, line 34: Because it outputs a hidden input, who breaks the display flex of the whole form.
So I need to hide them to only display the inputs we can modify.

skjnldsv:
closed in favor of #480

Issue #324, opened 2016-03-31
Distinction between first and last name


aspdye:

jancborchardt:
We should definitely try to do this in the backend. It’s already painful to need to put in the address in 5 or 6 different fields (everyone _HATES_ forms) so we definitely should not have separate input fields for the name. The old Contacts app didn’t have it either, at least not by default.

cc @tanghus @jbtbnl @babelouest if you feel like contributing again. :)

DeepDiver1975:
Name editing in the old contacts app

![bildschirmfoto von 2016-03-31 11-43-27](https://cloud.githubusercontent.com/assets/1005065/14171922/d1e8b346-f735-11e5-9f84-e0dcb98e3480.png)

DeepDiver1975:
editing addresses in old contacts app

![bildschirmfoto von 2016-03-31 11-44-31](https://cloud.githubusercontent.com/assets/1005065/14171948/fc19a7ce-f735-11e5-8ce2-dcf14067923d.png)

jancborchardt:
Damn … then it was probably wishful thinking, or it of course works like that on Android etc. ;)

skjnldsv:
In Android I think the contact app split automatically the Name you enter and divide the data into the FN field.

Example, if you type "John Doe", it will set John as first name and Doe as last name.
Maybe we should do that with a div expand possibility to correct the FN fields?

ionum:
Currently new Contact only have FN-Value (Displayname) and N-Value (Full Name with Family Name, Given Name, Additional Names, Honorific Prefixes, and Honorific Suffixes as in RFC 2426) is not editable.

ionum:
A lot of progams use algorithms to split names (eg. Outlook, Android, iOS), but none works perfectly. there are too much ways to write a full name....

aspdye:
#265

nWRqenkF:
related to #276 ??

nWRqenkF:
@ionum Somehow we already have an algorithm: "do nothing". Which is really annoying when using any other contact application.
Personally I would prefer the 8.1 contacts behaviour:
- Show only a single input field and try to split the input to first name and given name with an algorithm
- Give the user the possibility to display and edit the name detail fields, if he needs to correct the result of the algorithm

XZS1cpXn:
[...]

PR #418, opened 2016-05-29
Gui switch FN to N in Contact-Details

Rello:
as the automated FN splitting for First & Last Name is not working, the input field should be split into 2 separate ones.
They update the N:[0] and N:[1] directly

_...its my first Pull-Try. thank you for your patience :-)_

### DIFF
--- a/js/components/contactDetails/contactDetails_controller.js
+++ b/js/components/contactDetails/contactDetails_controller.js
@@ -19,6 +19,8 @@ angular.module('contactsApp')
 	ctrl.t = {
 		noContacts : t('contacts', 'No contacts in here'),
 		placeholderName : t('contacts', 'Name'),
+		placeholderLastName : t('contacts', 'Last Name'),
+		placeholderFirstName : t('contacts', 'First Name'),
 		placeholderOrg : t('contacts', 'Organization'),
 		placeholderTitle : t('contacts', 'Title'),
 		selectField : t('contacts', 'Add field ...')
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -42,6 +42,52 @@ angular.module('contactsApp')
 				}
 			},
 
+			lastName: function(value) {
+				var property = this.getProperty('n');
+				if (angular.isDefined(value)) {
+					var val = value;
+					// setter
+					if(property && Array.isArray(property.value)) {
+						val = property.value;
+						val[0] = value;
+					}
+					return this.setProperty('n', { value: val });
+				} else {
+					// getter
+					if(property) {
+						if (Array.isArray(property.value)) {
+							return property.value[0];
+						}
+						return property.value;
+					} else {
+						return undefined;
+					}
+				}
+			},
+
+			firstName: function(value) {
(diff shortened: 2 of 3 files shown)


z000ao8q:
0a7d93b

DeepDiver1975:
N is not first name and last name - see https://tools.ietf.org/html/rfc6350#section-6.2.2

Rello:
sorry - I updated the typo in the first Post.
N is an array with 
N:[0] = Last
N:[1] = First

...and others. but I think to go for the workaround Last and First are the most important ones.
Currently it does not work at all (only updating FD and not N) together with the sync clients as iOS as they rely on N

DeepDiver1975:
What happens to N:[2+] of existing vcards while editing?
Dataloss is not accepable.

Rello:
they are untouched because the specific index is used in the setter in contact_model.js
`val[0] = value;`

if feels like a workaround until someone has a better idea. but currently the whole N\* is ignored completely when editing via the GUI.
As only the FN is updated in the GUI, you make changes which don´t reach e.g. iOS or Mac OS Calender, as the main field (n*) remains on the old content

with the change at least first & last can be edited again in GUI - better than none at all

Rello:
existing contact:
screen shot 2016-05-31 at 22 01 52

new contact/empty input Form
screen shot 2016-05-31 at 22 02 49

alternatively they could be merged into one line also

I changed it locally on my install and it looks fine.
any other suggestion on where we go from here?

An other idea would be to look at the section where the current FN input is processed and to look at derivation rules to get the N out of this.
But as the Display Name (FN) can have any input - and the N is a strict given structure - I think N->FN makes more sense compared to FN->N

skjnldsv:
I don't like the Idea of multiple inputs! Can't you automatic split the input into two value? Like android? :)

DeepDiver1975:
Splitting fn to N is not easy. Different cultures have different ways to split. The android code which does the magic is hugh - i will link the code later.

skjnldsv:
https://android.googlesource.com/platform/frameworks/opt/vcard/+/master/java/com/android/vcard/VCardBuilder.java ?

DeepDiver1975:
https://github.com/android/platform_packages_providers_contactsprovider/blob/master/src/com/android/providers/contacts/NameSplitter.java

skjnldsv:
Much better! :D

phsc84:
If you would implement it like this https://github.com/owncloud/contacts/issues/324#issuecomment-207950261, the algorithm to split the names would not necessarily need to be as complex as the android code, because the user still had the possibility to correct wrongly splitted names.
Furthermore the automatic splitting could be disabled, if the N fields were already filled. That way, the risk to overwrite existing values would be zero.

jancborchardt:
Regarding the layout: Let’s please put the input fields for first and last name next to each other, not one below the other. That just looks weird. Since the first name has variable length, we should do that via Javascript just like we do for the Organization before the Title field. :)

DeepDiver1975:
I'm closing this for now - we have added a separate N field for now - #437

PR #448, opened 2016-07-06
Spelling mistake

napobear:


DeepDiver1975:
needs to be fixed on transifex - https://www.transifex.com/owncloud-org/owncloud/translate/#it/contacts/76888704

Issue #449, opened 2016-07-09
Display number of contacts below the contacts list


TecJon:
### Display total number of contacts

I suggest to display the total number of contacts below the contact list for convenience and verification. 

I imported a .vcf file with several hundred contacts and couldn't verify if all contacts got imported correctly. In fact they didn't and I only found out by chance that contacts were missing.

![numberofcontacts](https://cloud.githubusercontent.com/assets/11375752/16707485/d42d7c48-45cf-11e6-92eb-cfd1fff5f1cf.png)

DeepDiver1975:
fixed with #453

PR #453, opened 2016-07-16
Add number of displayed contacts to the bottom of the contact list

juliusknorr:
as described in #449

### DIFF
--- a/templates/contactList.html
+++ b/templates/contactList.html
@@ -2,7 +2,7 @@
         <button ng-show="!ctrl.loading && !ctrl.invalid" class="app-content-list-button" id="new-contact-button"
 				type="button" name="button" ng-click="ctrl.createContact()">{{ctrl.t.addContact}}</button>
         <div class="app-content-list-item"
-			 ng-repeat="contact in ctrl.contactList = (ctrl.contacts | contactGroupFilter:ctrl.routeParams.gid | localeOrderBy:'displayName' | filter:query) track by contact.uid()"
+			 ng-repeat="contact in ctrl.contactList = (ctrl.contacts | contactGroupFilter:ctrl.routeParams.gid | localeOrderBy:'displayName' | filter:query) as filtered track by contact.uid()"
 			 contact data="contact"
 			 ng-click="setSelected(contact.uid())"
 			 ng-class="{active: contact.uid() === ctrl.getSelectedId()}">
@@ -12,5 +12,8 @@
 		        <div class="icon-search"></div>
 			<h2>{{ctrl.t.emptySearch}}</h2>
 		</div>
+	</div>
+	<div ng-if="filtered.length>0" class="contacts-list-count">
+		{{ getCountString(filtered) }}
 	</div>
 </div>
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,6 +29,7 @@
 		"_": false,
 		"OC": false,
 		"t": false,
+		"n": false,
 		"dav": false,
 		"OCA": false
 	}
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -472,6 +472,16 @@ ul.addressBook-share-list li {
 	opacity: 1;
 }
 
+/* styling for contact count in contact list */
+.contacts-list-count {
+	text-align:center;
+	color: #333;
+	padding:25px;
+	position: relative;
+	height: 68px;
+	border-top: 1px solid #eee;
+}
+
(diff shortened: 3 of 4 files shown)


jancborchardt:
Ideally we should do it like in the old Contacts app before and display the number of contacts in the sidebar in the respective row of the group or all contacts.

As far as I remember this wasn’t possible due to some technical reason cc @Henni @DeepDiver1975 – but if it’s possible now then we should definitely place it there.

DeepDiver1975:
> but if it’s possible now then we should definitely place it there.

it is possible now to display the number of contacts - as soon as we switch over to lazy loading (which is necessary for handling bigger addressbooks) we no longer get this information from the server.

DeepDiver1975:
we can merge this now but it will be a feature regression as soon as we switch to lazy loading

templates/contactList.html
```diff
@@ -13,4 +13,7 @@
 			<h2>{{ctrl.t.emptySearch}}</h2>
 		</div>
 	</div>
+	<div ng-if="filtered.length>0" class="contacts-list-count">
+		{{ filtered.length }} {{ filtered.length==1 ? ctrl.t.oneContact : ctrl.t.multipleContacts }}
```
Henni: @DeepDiver1975 I believe this way of translation might create problems. Wasn't there something about different plurals in other languages?

Henni:
Besides my comment, I'm voting for merging this, even though it will create a regression in the future.
This PR helps us discover problems with the import functionality.

templates/contactList.html
DeepDiver1975: absolutly - this will not work out!

Henni: @DeepDiver1975 do you remember how/where we fixed this before?
@juliushaertl Please take another look. We would be happy to merge this PR after this is resolved!

juliusknorr: Yes, will have a look. If anyone has a hint how to solve the plural translations, that would be great. Or is it basically just moving to translation strings with placeholders like "%d contact"/"%d contacts"

jancborchardt:
Ok, sounds fine for me. We should really move that to the sidebar in the future though.

templates/contactList.html
DeepDiver1975: in pure js we would call n() instead of t() - but with angular this is a bit different.

@raimund-schluessler @georgehrke what is your approach with respect to plural strings in angular and ownCloud? THX

raimund-schluessler: I simply use n(), see:
https://github.com/owncloud/tasks/blob/master/js/public/app.js#L1093

Is there a drawback of doing so?

DeepDiver1975: > I simply use n(), see:
> https://github.com/owncloud/tasks/blob/master/js/public/app.js#L1093
> 
> Is there a drawback of doing so?

where in the template is it used then?

No drawback - I'm just asking to gain knowledge :wink:

georgehrke: I use t (and n) as well.

In this case I would write a filter to display the different strings.

georgehrke: see https://github.com/owncloud/calendar/blob/master/js/app/filters/importerrorfilter.js

raimund-schluessler: > where in the template is it used then?

It is used here:
https://github.com/owncloud/tasks/blob/master/templates/part.tasklist.php#L25

Using this as a filter may be more clear (and easier to test? :innocent:)

juliusknorr:
I've just pushed an update. Please check again.

Henni:
Code looks good. Still have to test it.

DeepDiver1975:
eslint is failing:

```
[19:39:53] Using gulpfile ~/build/owncloud/contacts/gulpfile.js

[19:39:53] Starting 'eslint'...

[19:39:55] 

/home/travis/build/owncloud/contacts/js/components/contactList/contactList_controller.js

  18:10  error  'n' is not defined  no-undef

/home/travis/build/owncloud/contacts/js/services/contact_service.js

   40:10  warning  Unexpected console statement        no-console

  109:4   warning  Unexpected console statement        no-console

  129:46  warning  Don't make functions within a loop  no-loop-func

✖ 4 problems (1 error, 3 warnings)
```

Henni:
@DeepDiver1975 yeah right...
We have to tell eslint which globals exist.
`n` has to be added to `.eslintrc.json`

juliusknorr:
Ah forgot to commit that as well. Added it.

DeepDiver1975:
:+1:

PR #472, opened 2016-07-29
💀 Deathdate … and anniversary! 🎉

DeepDiver1975:


jancborchardt:
_Not sure if I should be worried about you or me now …_

js/services/vCardProperties.js
```diff
@@ -66,6 +66,14 @@ angular.module('contactsApp')
 			readableName: t('contacts', 'Birthday'),
 			template: 'date'
 		},
+		anniversary: {
+			readableName: t('contacts', 'Anniversary'),
+			template: 'date'
+		},
+		deathdate: {
+			readableName: t('contacts', 'Date of Death'),
```
jancborchardt, line 74: Correct capitalization should be »Date of death«, only first character capitalized.

DeepDiver1975:
> Not sure if I should be worried about you or me now …

no worry this introduces anniversary as well ....

js/services/vCardProperties.js
DeepDiver1975, line 74: please fix right away .... thx

DeepDiver1975:
sync to calendar is in core - https://github.com/owncloud/core/pull/25655

jancborchardt:
So @ChristophWurst noted something: Does it really make sense to save date of death for a person in Contacts? Cause … you will not contact them anymore really.

I guess it makes sense to remember someone though, or just for completeness … I just wanna make sure we’re not just stupidly implementing the spec, because well, this in particular is a very slim usecase. ;)

spoorun:
@jancborchardt A good point. We had thought of that, and from a personal perspective, I retain contacts who have died, often for a memory or commemoration - for many people it is an important a date to remember someone - as important as a birthday is of someone still alive. So it's not a huge use case, but certainly not a 'very slim' one. As people grow older, alas that usecase also grows.

jancborchardt:
Ok, just wanted to make sure it’s a conscious decision adding this, not just cause you’re in the process of adding anniversary anyway. :)

MariusBluem:
Our satanistic users will like that :smiling_imp:

spoorun:
I guess there's nothing preventing future events being recorded either.. for those who are exceptionally well-planned!

PR #491, opened 2016-08-17
add viewBox to app icon for proper scaling in firefox

MorrisJobke:
- this already works fine with Chrome, because there the height and width are used by default
- Firefox needs a viewBox specified to scale at all
- also removes unused XML attributes from the SVG
- ref nextcloud/server#905

cc @LukasReschke @jancborchardt


MorrisJobke:
@DeepDiver1975 @Henni @juliushaertl and @irgendwie for the most recent contributors :)

PR #493, opened 2016-08-22
Move button to the side bar - fixes #483

### LINKED ISSUE #483: Move new Contacts button to sidebar
According to the removal of the **Search for new Messages**-Button in the Mail app, I think we should move the **New Contacts** Button to the left sidebar, because we also have the **New Message** Button in the Mail-app at this place 😉 
### Mail:

mail_repro
### Contacts:

contacts_repro

To adjust it a bit better, we could also call it **\+ New Contact** instead of **New Contact**

cc @jancborchardt

DeepDiver1975:


irgendwie:
Works :+1:

Issue #430, opened 2016-06-03
Contacts 1.3.1 - Doesn't load (spinning wheel)


Akhenaton:
### Steps to reproduce
1. update from 1.3 to 1.3.1
2. Do some sync with with KDE's Kaddressbook
3. Open in the web UI the app
### Expected behaviour

Contacts should appear.
### Actual behaviour

Spinning wheel and never loads.
### Server configuration

Linux
Apache2
Mysql
PHP 7.0.7
ownCloud 9.0.2
Contacts version 1.3.1

**List of activated apps:**
Enabled:
- activity: 2.2.1
- announcementcenter: 1.1.2
- calendar: 1.2.2
- comments: 0.2
- contacts: 1.3.1.0
- dav: 0.1.6
- documents: 0.12.0
- external: 1.2
- federatedfilesharing: 0.1.0
- federation: 0.0.4
- files: 1.4.4
- files_pdfviewer: 0.8.1
- files_sharing: 0.9.1
- files_texteditor: 2.1
- files_trashbin: 0.8.0
- files_versions: 1.2.0
- files_videoplayer: 0.9.8
- firstrunwizard: 1.1
- gallery: 14.5.0
- music: true
- notifications: 0.2.3
- provisioning_api: 0.4.1
- systemtags: 0.2
- templateeditor: 0.1
- updatenotification: 0.1.0
#### ownCloud log (data/owncloud.log)

[...]

PR #502, opened 2016-08-29
Fix loading spinner with usernames containing spaces.

irgendwie:
This fix should solve the issue with blanks in OC/NC usernames:
 #430 - @AndrzejMM @ossadmirer314 @JSoko @DeepDiver1975 

The bug is caused by not escaping urls in the dav lib.

This is a temporary fix, since we are switching to another dav lib anyway: #216


JSoko:
Yes, it works! Thanks a lot! :-)

MariusBluem:
:+1:

PR #509, opened 2016-09-03
Update dependencies and increase version number

irgendwie:
Update version string for the next version and the dependencies...

Do **not** merge - testing if everything still works

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@
 
 ![codecov.io](https://codecov.io/github/owncloud/contacts/branch.svg?branch=master)
 
-**A contacts app for [ownCloud](https://owncloud.org).**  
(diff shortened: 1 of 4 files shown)


irgendwie:
- Update to README to show compatibility with ownCloud and Nextcloud.
- Update dependencies
- Everything seems to still work

Review please! @Henni @DeepDiver1975 @skjnldsv

DeepDiver1975:
:+1:

enoch85:
Please try to use the same version number as the latest release as this will cause trouble for me building the VM as it gets the latest version from your info.xml file: https://github.com/nextcloud/vm/blob/master/nextcloud_install_production.sh#L481

Thanks!

PR #553, opened 2016-12-08
Date input enhancements

### LINKED ISSUE #122: Support text based birthdays
[RFC 6350](https://tools.ietf.org/html/rfc6350#section-6.2.5) allows vCards entries like `BDAY;VALUE=text:circa 1800` which should be supported as well as date birthdays

DeepDiver1975:
* Allow selected year to be smaller then 1000
* Allow text only input for dates - e.g. ca 1802

fixes #122


tomneedham:
it works 🎉 👍

PR #557, opened 2016-12-14
Add avatar download and remove

DeepDiver1975:
Signed-off-by: John Molakvoæ (skjnldsv) <skjnldsv@protonmail.com>

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -71,18 +71,18 @@
 	opacity: .8;
 }
 
-.contactdetails__header .icon-delete-white,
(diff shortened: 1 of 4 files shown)


tomneedham:
Hmm now with this PR, on mobile / touch the UX is kind of broken since there are three click areas on the icon that you cant see without hover. Before you could tap wherever and it would bring up the upload dialog.

Can we just show the upload one on mobile and leave the other features for desktop?

screen shot 2016-12-14 at 21 13 13

tomneedham:
+ icon-fullscreen-white is missing

PR #569, opened 2017-01-25
Added sort functionality (visible in settings, bottom left)

### LINKED ISSUE #208: Contact sort/display order missing
### Expected behaviour

There should be a way to sort the contact list/change the display order (like it was on oc 8.2 contacts app).

kontakte_owncloud2

I was told this wasn't implemented yet, so no further information, I just created this ticket as a backlog.

snaiperskaya96:
Closes #208

### DIFF
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -30,6 +30,18 @@ angular.module('contactsApp')
 				}
 			},
 
+			sortFirstName: function() {
+				return [this.firstName(), this.lastName()];
+			},
+
+			sortLastName: function() {
+				return [this.lastName(), this.firstName()];
+			},
+
+			sortDisplayName: function() {
+				return this.displayName();
+			},
+
 			displayName: function() {
 				return this.fullName() || this.org() || '';
 			},
@@ -39,7 +51,7 @@ angular.module('contactsApp')
 				if (property) {
 					return property.value[1];
 				} else {
-					return this.fullName();
+					return this.displayName();
 				}
 			},
 
@@ -48,7 +60,16 @@ angular.module('contactsApp')
 				if (property) {
 					return property.value[0];
 				} else {
-					return this.fullName();
+					return this.displayName();
+				}
+			},
+
+			additionalNames: function() {
+				var property = this.getProperty('n');
+				if (property) {
+					return property.value[2];
+				} else {
+					return '';
 				}
 			},
 
--- a/js/models/contact_model.js
+++ b/js/models/contact_model.js
@@ -34,6 +34,24 @@ angular.module('contactsApp')
 				return this.fullName() || this.org() || '';
 			},
 
+			firstName: function() {
+				var property = this.getProperty('n');
+				if (property) {
+					return property.value[1];
+				} else {
+					return '';
+				}
+			},
+
+			lastName: function() {
+				var property = this.getProperty('n');
+				if (property) {
+					return property.value[0];
+				} else {
+					return '';
+				}
+			},
+
 			fullName: function(value) {
 				var model = this;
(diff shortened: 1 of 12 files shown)


DeepDiver1975:
@snaiperskaya96 Thanks a lot - I'll review this asap!

snaiperskaya96:
I'm a bit confused about codecov checks, never heard/used it, can you please explain what's the problem with it?

DeepDiver1975:
> what's the problem with it?

The basic problem is that you did add code which is no covered by unit tests. As a result the number of covered lines goes down. This is why the check fails.

snaiperskaya96:
Oh i see, so I should add some unit tests?

DeepDiver1975:
> I should add some unit tests?

that would be awesome - yes - THX

DeepDiver1975:
@snaiperskaya96 thanks for the tests - I'll have a look at this tomorrow

snaiperskaya96:
Sorry about the silly tests, didn't really know what else to try :stuck_out_tongue:

DeepDiver1975:
In case the chosen sort criteria is not set within one contact I'd like to see a fallback to FN.
Furthermore the sorting by displayname should be added as this is the 'old' behavior.

THX

snaiperskaya96:
@DeepDiver1975 
>In case the chosen sort criteria is not set within one contact I'd like to see a fallback to FN.
Furthermore the sorting by displayname should be added as this is the 'old' behavior.

Done 😄

js/models/contact_model.js
```diff
@@ -34,6 +34,24 @@ angular.module('contactsApp')
 				return this.fullName() || this.org() || '';
 			},
 
+			firstName: function() {
+				var property = this.getProperty('n');
+				if (property) {
+					return property.value[1];
```
DeepDiver1975, line 52: From my pov we need to do this differently.
1. in case there are two contacts with the same first name we need to use the last name (or the additional first names???) as second sort criteria
2. in case no last name or last name is set we need to sort by displayname otherwise this will be chaotic
3. we should think about to adjust the name which is displayed: 
   - by last name: last name + first name + additional names
   - by first name: first name + additional names + last name

4. Open question: what shall we do with those contacts which have no N but only a FN - e.g. companies have that

snaiperskaya96: > Open question: what shall we do with those contacts which have no N but only a FN - e.g. companies have that

Guess i have no control over these

DeepDiver1975:
let's try the folllowing:
1. keep sorting based on displayname
2. displayname changes based on sort criteria
3. displayname is built based on sort criteria:
 - default sort: displayname is FN - if FN is not set: use N (basically just like the code does today)
 - first name sort: displayname is N[1] + N[2] +  N[0] - if N is not set: use FN
 - last name sort: displayname is N[0] + ',' + N[1] + N[2] - if N is not set: use FN
 - email sort: I'd drop that - there are multiple emails in one contact

snaiperskaya96:
Uops, sorry, did some changes before reading your last comment, have a look if it might be okay now.
Basically is now considering both first and last name while sorting (if the sort criteria is equal in both the entities), the display name changes based on the sort criteria and dropped the email sorting

DeepDiver1975:
- [x] there is no sort preselected on first start

snaiperskaya96:
> there is no sort preselected on first start

Actually there is, but i changed keys naming so maybe you got the old values cached?
Try to delete this row from localStorage 
screen shot 2017-02-01 at 7 28 28 am

By default there should be none, and the default selected one should be Display Name. Then as soon as you select another one it should create a row into the localStorage and reference it

DeepDiver1975:
Thanks @snaiperskaya96 !

PR #600, opened 2017-06-30
Split contact in two columns on large screens

### LINKED ISSUE #354: Split Contact Details into two colums

william-weber:
Fixes #354 
* add 'addresses' and 'nonAddresses' to contact model
* update those two on add/remove property
* show addresses and nonAddresses in own divs
* use CSS 'float' on divs when display width >=1450 for two column layout

Critique welcome (esp on style) - this is my first real work with angular.

### DIFF
--- a/css/public/style.css
+++ b/css/public/style.css
@@ -155,15 +155,31 @@ avatar.avatar--missing .avatar-upload {
 	object-fit: cover;
 }
 
-.contactdetails__header + section {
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
awesome job as well - thanks a lot! 🎉

PR #606, opened 2017-07-28
Update summary and description Contacts App

tobiherr:
@pmaier1

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -2,10 +2,11 @@
 <info>
     <id>contacts</id>
     <name>Contacts</name>
-    <summary>A contacts app for ownCloud</summary>
+    <summary>Manage and sync your contacts across multiple devices</summary>
     <description>
-The ownCloud contacts app is a user interface for ownCloud's CardDAV server.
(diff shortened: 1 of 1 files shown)


appinfo/info.xml
```diff
@@ -2,10 +2,11 @@
 <info>
     <id>contacts</id>
     <name>Contacts</name>
-    <summary>A contacts app for ownCloud</summary>
+    <summary>Manage and sync your contacts across multiple devices</summary>
```
DeepDiver1975, line 5: the contacts app has nothing to do with sync and mutliple devices - this is a backend functionality provided by core.

DeepDiver1975 (changes requested):

pmaier1 (changes requested):
Change category to 'productivity', please.

appinfo/info.xml
pmaier1: That's what I said but marketing does not want to describe the app technically but the use case.. IMO this is misleading and confusing for sysadmins but I'm more than done discussing this. People want to magically make our apps more than they are because it looks better.

DeepDiver1975: I will not accept this change then

PR #634, opened 2018-02-02
Now it's possible to star contacts

Julian1998:
Also I fixed a little tooltip error when hovering about the delte icon

### DIFF
--- a/js/components/contactDetails/contactDetails_controller.js
+++ b/js/components/contactDetails/contactDetails_controller.js
@@ -23,13 +23,25 @@ angular.module('contactsApp')
 		placeholderTitle : t('contacts', 'Title'),
 		selectField : t('contacts', 'Add field ...'),
 		download : t('contacts', 'Download'),
-		delete : t('contacts', 'Delete')
+		delete : t('contacts', 'Delete'),
+		favour : t('contacts', 'Favour'),
+		favourites : t('contacts', 'Favourites')
 	};
 
 	ctrl.fieldDefinitions = vCardPropertiesService.fieldDefinitions;
 	ctrl.focus = undefined;
 	ctrl.field = undefined;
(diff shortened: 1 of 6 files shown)


Julian1998:
@DeepDiver1975 Can you explain me why codecov is failing?

js/components/contactDetails/contactDetails_controller.js
```diff
@@ -23,14 +23,26 @@ angular.module('contactsApp')
 		placeholderTitle : t('contacts', 'Title'),
 		selectField : t('contacts', 'Add field ...'),
 		download : t('contacts', 'Download'),
-		delete : t('contacts', 'Delete')
+		delete : t('contacts', 'Delete'),
+		favour : t('contacts', 'Favour'),
```
DeepDiver1975, line 27: -u

templates/contactDetails.html
```diff
@@ -24,8 +24,10 @@ <h2>
 				</div>
 			</div>
 			<div id="details-actions">
+				<button ng-show="!ctrl.starred" ng-click="ctrl.starContact()" class="icon-star" title="{{ctrl.t.favour}}"></button>
```
DeepDiver1975, line 27: two buttons? why not change the class inside.

DeepDiver1975 (changes requested):

templates/contactDetails.html
Julian1998: Didn't thought about that! Good point

js/components/contactDetails/contactDetails_controller.js
Julian1998: -u? What do you mean by that

DeepDiver1975:
:coffin:

PR #668, opened 2019-08-29
Use drone convert to get a new drone matrix

phil-davis:


phil-davis:
Hmmm - this passes but my attempt with starlark does not. Need to look harder at the differences tomorrow.

phil-davis:
works

Issue #680, opened 2020-01-14
JS tests failing in CI


phil-davis:
About 50% of nights, the nightly CI javascript-tests pipeline is failing.
e.g. https://drone.owncloud.com/owncloud/contacts/213/4/6
```
14 01 2020 02:15:49.424:ERROR [launcher]: Cannot start firefox via Remote WebDriver
	null
14 01 2020 02:15:49.424:ERROR [launcher]: firefox via Remote WebDriver stdout: 
14 01 2020 02:15:49.424:ERROR [launcher]: firefox via Remote WebDriver stderr: 
14 01 2020 02:15:49.424:ERROR [launcher]: firefox via Remote WebDriver failed 2 times (cannot start). Giving up.
```

phil-davis:
Failed again last night: https://drone.owncloud.com/owncloud/contacts/236/4/6

phil-davis:
The karma dependencies have been updated in #682 
Now to wait and see the results of the nightly CI  for the next few nights.

phil-davis:
Last night passed: https://drone.owncloud.com/owncloud/contacts/242
That's a good start :)

phil-davis:
CI has been good the last week. Probably PR #682 has fixed this issue. Closing.

PR #681, opened 2020-01-14
Add sleep before JS tests

phil-davis:
to see if the intermittent JS  tests failures are just a startup timing issue

issue #680

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -23,6 +23,9 @@ config = {
 
 	'javascript': {
 		'coverage': True,
+		'extraCommandsBeforeTestRun': [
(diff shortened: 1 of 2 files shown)


phil-davis:
When I restart the nightly drone job it is always passing. And when I restart this drone job it is also always passing. To really be able to test this I need to be able to ensure that the drone pipeline starts on a new drone agent - because it is likely that the timing issue is only going to be a problem onn a new agent that needs to actually fetch ever docker... image from across the network.

phil-davis:
Nightly results are better since PR #682 
No need for a timing hack like this.

Issue #693, opened 2020-06-17
javascript tests failing in CI


phil-davis:
https://drone.owncloud.com/owncloud/contacts/427/5/6

The Javascript tests are failing last night in CI. I tried restarting a few times and it still failed.

Needs investigation. Did some version of some test dependency change yesterday or?

phil-davis:
`17 06 2020 03:19:01.269:ERROR [launcher]: Cannot start firefox via Remote WebDriver`

Maybe some new firefox release or?

PR #694, opened 2020-06-17
Use selenium version 3.141.59-20200525

jasson99:
This PR reverts the selenium version to 3.141.59-20200525, since the javascript tests failed on the latest version.

## Related Issue
- https://github.com/owncloud/contacts/issues/693


phil-davis (approved):
It's a bit sad to have to do stufff like this, but nothing we can do about it.

PR #723, opened 2021-09-04
Update .drone.star and drop PHP 7.2

phil-davis:
Part of issue https://github.com/owncloud/core/issues/39134
and https://github.com/owncloud/QA/issues/687

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -15,15 +15,7 @@ config = {
         "master",
     ],
     "appInstallCommand": "make npm",
-    "codestyle": {
-        "ordinary": {
-            "phpVersions": [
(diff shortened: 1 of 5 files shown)


mmattel:
Regarding the failing CI, seems that composer is not installed:
`Makefile:44: *** composer is not available on your system, please install composer.  Stop.`

phil-davis:
I removed the check from `Makefile` - that check is often annoying, and specially now with steps that use `Makefile` targets that do not require PHP or `composer`. It has been removed in other repos, so I need to remember to do it in each repo that I update!

dpakach (approved):

PR #745, opened 2022-11-25
[tests-only] [full-ci] Run phan with PHP 7.3

phil-davis:
`phan` will report errors if any of the code is not compatible with PHP 7.3.

This app still supports PHP 7.3 when used with oC10 core 10.11 and earlier.


phil-davis:
https://drone.owncloud.com/owncloud/contacts/1488/5/3
`phan` passed with PHP 7.3

kiranparajuli589 (approved):

PR #757, opened 2026-04-08
feat: add translation sync workflow

DeepDiver1975:
## Summary

Adds a nightly translation sync workflow that calls the reusable workflow from [owncloud/reusable-workflows](https://github.com/owncloud/reusable-workflows) (owncloud/reusable-workflows#24).

Push is authenticated via the repo's own `GITHUB_TOKEN` — no PAT or SSH key needed.

`TX_TOKEN` must be available as a repository or organisation secret.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/translation-sync.yml
+++ b/.github/workflows/translation-sync.yml
@@ -5,6 +5,8 @@ on:
     - cron: '0 2 * * 1'
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   sync:
     uses: owncloud/reusable-workflows/.github/workflows/translation-sync.yml@main
--- a/.github/workflows/translation-sync.yml
+++ b/.github/workflows/translation-sync.yml
@@ -2,7 +2,7 @@ name: Translation Sync
 
 on:
   schedule:
-    - cron: '0 2 * * 1'
+    - cron: '0 2 * * *'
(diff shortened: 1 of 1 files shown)


mmattel (approved):

jvillafanez:
I have a question: it seems the workflow "read + push" and then "pull + write". I think it's better to split this because we want frequent pushes, so the new strings are quickly available for the translators.
With the current cron, the job will run every Monday. If a write a new string on Tuesday, I'll have to wait until Monday (up to 7 days) for the workflow to run AND another 7 days for the next Monday to get the translation because the string won't be translated immediately.

Assuming we want a weekly schedule, it might be better to push the strings everyday so the translators have time to translate during the week. We can pull the translated strings weekly if we want.
In addition, if we want the weekly schedule, we might need another trigger to pull translations for "emergencies". Usually, right before releasing, we might need to pull translations on demand.

DeepDiver1975:
> In addition, if we want the weekly schedule, we might need another trigger to pull translations for "emergencies". Usually, right before releasing, we might need to pull translations on demand.

the workflows in the individual repos can be triggered manually

DeepDiver1975:
point taken on the weekly update ....

PR #764, opened 2026-04-22
chore(ci): move PHP integration tests to GitHub workflow

phil-davis:


phil-davis:
https://github.com/owncloud/contacts/actions/runs/24761746096/job/72446858376?pr=764

PHP Integration test ran. There is only 1!

Hmmm - when I change the name of the job from "PHP Unit" to "PHP Unit and Integration tests" then the expected result "PHP Unit" does not happen. So that needs to be changed in the "expected result" settings for the repo.

Edit: fixed by changing the text back to just "PHP Unit". We will understand that "PHP Unit" includes possible tests categorised as "PHP Integration".

phil-davis:
Note: there are still Javascript tests to move to GitHub workflows.
In drone, they have a selenium browser defined, so need to look at that in a PR after this to see if it is really needed like that, and maybe make the js-unit workflow more flexible.

.github/workflows/main.yml
```diff
@@ -47,10 +47,11 @@ jobs:
       php-versions: ${{ needs.get-vars.outputs.php-versions }}
 
   php-unit:
-    name: PHP Unit
+    name: PHP Unit and Integration tests
```
DeepDiver1975: just leave the name as 'PHP Unit' and the branch protection will work unchanged.

DeepDiver1975 (approved):

